PR #2, opened 2019-10-27
changed secrets naming

adaptive:
Preparing for large Github Action operations where many secrets are stored, and naming gets confusing.
`CLOUDFLARE_API_KEY`	
`CLOUDFLARE_EMAIL`

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,15 +23,15 @@ jobs:
       - name: Publish
         uses: cloudflare/wrangler-action@1.0.0
         with:
-          apiKey: ${{ secrets.apiKey }}
-          email: ${{ secrets.email }}
+          apiKey: ${{ secrets.CLOUDFLARE_API_KEY }}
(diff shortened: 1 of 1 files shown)


ashleygwilliams:
hey @adaptive -a few things here:

1. i think if we were to make this change, we'd want it to be inline with the env vars that wrangler takes, so CF_API_KEY, and CF_EMAIL
2. we'd need to update the README!
3. i think this would be a breaking change so we'd need to consider how to version/communicate that!

thanks for this! as for what next  steps are, i defer to @signalnerve :)

adaptive:
CF_API_KEY and CF_EMAIL are fine, the most important is to differentiate between secrets when using from different providers e.g. AWS, IBM, Cloudflare...

Initially, I thought of changing the whole action but made no sense, I just changed the README, so that wouldn't break the Github action.

PR #7, opened 2019-11-07
Add workingDirectory option for publishing a subdirectory

fshot:
This addresses https://github.com/cloudflare/wrangler-action/issues/3

Usage is as suggested in the issue:

```
- name: Publish worker one
  uses: cloudflare/wrangler-action@1.0.0
  with:
    apiKey: ${{ secrets.CF_API_KEY }}
    email: ${{ secrets.CF_EMAIL }}
    workingDirectory: 'worker-one'
- name: Publish worker two
  uses: cloudflare/wrangler-action@1.0.0
  with:
    apiKey: ${{ secrets.CF_API_KEY }}
    email: ${{ secrets.CF_EMAIL }}
    workingDirectory: 'worker-two'
```

PR #11, opened 2019-11-22
📡 1.1.0

signalnerve:
It's wrangler-action 1.1.0! 📡

- Support `apiToken` as an authentication method
  - Deprecates `apiKey` and `email`, making them optional parameters and encouraging `apiToken` usage in build logs
- Support `wranglerVersion` for installing a specific Wrangler version for your build
- Per #7, support for `workingDirectory` to run `wrangler-action` in a specific directory in your repo
- Adds a test Workers project under the `test` directory. This is used in the repo's new set of workflows (see below)
- ~Adds a GitHub Action workflow that:~
  - ~Lints `entrypoint.sh` to ensure that the shell script looks correct~
  - ~Runs the action with various config options to ensure future pushes don't introduce regressions~
- Our GitHub Actions stuff is busted right now, so I've added the above item but left the action disabled, under `.github/actions/deploy.yml.disabled`. When/if we resolve that, I'll rename it back to ` .yml`


signalnerve:
Our GitHub Actions stuff across the org is busted, so here's a series of tests to confirm that this stuff all works as expected. Note that the final failing test is intentional, to make sure that we get the correct error message back when no auth has been provided.

https://github.com/signalnerve/static-test/commit/b49842ea236d3ff5ca6cf07c36b01f5b99531332/checks?check_suite_id=324496997

And the accompanying inputs for that action:

https://github.com/signalnerve/static-test/blob/b49842ea236d3ff5ca6cf07c36b01f5b99531332/.github/workflows/deploy.yml

minddust:
Yes! thanks for the token support :)

README.md
```diff
@@ -21,50 +21,76 @@ jobs:
...
   deploy:
     # ... previous configuration ...
     steps:
-      uses: cloudflare/wrangler-action@1.0.0
+      uses: cloudflare/wrangler-action@1.1.0
       with:
-        # ... api key and email ...
+        # ... token ...
         environment: 'production'
 ```
```
EverlastingBugstopper: do you think it's worth just putting in a complete config here? people like copy + paste

action.yml
```diff
@@ -9,11 +9,13 @@ runs:
 inputs:
   apiKey:
     description: "Your Cloudflare API Key"
-    required: true
+  apiToken:
+    description: "(Legacy) Your Cloudflare API Token"
```
EverlastingBugstopper: API Key is legacy, not token, correct?

entrypoint.sh
```diff
@@ -6,33 +6,45 @@ export HOME="/github/workspace"
...
-  if [ -z "${1}" ]
-  then
-    >&2 echo "Unable to find ${2}. Did you add a GitHub secret called key ${2}, and pass in secrets.${2} in your workflow?"
-    exit 1
-  fi
-}
-
 curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.0/install.sh | bash
+
+# shellcheck source=/dev/null
```
EverlastingBugstopper: what does this do?

entrypoint.sh
```diff
@@ -6,33 +6,45 @@ export HOME="/github/workspace"
...
-npm i @cloudflare/wrangler -g
+if [ -z "$INPUT_WRANGLERVERSION" ]
+then
+  npm i @cloudflare/wrangler -g
+else
+  npm i "@cloudflare/wrangler@$INPUT_WRANGLERVERSION" -g
+fi
 
-if ! [ -z "$INPUT_WORKINGDIRECTORY" ]
+if [ -n "$INPUT_WORKINGDIRECTORY" ]
```
EverlastingBugstopper: what is -n vs -z?

EverlastingBugstopper: may be worth adding some comments.

action.yml
minddust: same. guess the api key was meant to be legacy?

entrypoint.sh
```diff
@@ -6,33 +6,45 @@ export HOME="/github/workspace"
...
+if [ -n "$INPUT_APIKEY" ] || [ -n "$INPUT_EMAIL" ]
+then
+  echo "Using legacy email and API key authentication"
+  export CF_EMAIL="$INPUT_EMAIL"
+  export CF_API_KEY="$INPUT_APIKEY"
+fi
 
-export CF_EMAIL="$INPUT_EMAIL"
-export CF_API_KEY="$INPUT_APIKEY"
+if [ -z "$INPUT_APITOKEN" ] && [ -z "$INPUT_EMAIL" ] && [ -z "$INPUT_APIKEY" ]
```
minddust: this doesn't fail if only one of (email, key) is given, which is insufficient

action.yml
signalnerve: oh oops! thank you, good catch

README.md
signalnerve: good idea, added!

entrypoint.sh
signalnerve: Added comments for each of these conditionals, thanks!

entrypoint.sh
signalnerve: nice catch! I rewrote this a bit to bail out if either api key or email are set, but not both. our CI is a little busted rn so you can see it working in action here: https://github.com/signalnerve/static-test/commit/e85ebdc37f6a9ea47d5c7dc1e191b74b4e007a3f/checks?check_suite_id=349537573

signalnerve:
@EverlastingBugstopper @minddust thanks for the review! Updated the branch w/ your suggestions, appreciate it :) 

our org-wide github actions is busted right now, so i ran this with another project to make sure all the new features are working, you can find it here: https://github.com/signalnerve/static-test/commit/e85ebdc37f6a9ea47d5c7dc1e191b74b4e007a3f/checks?check_suite_id=349537573

do you mind re-reviewing and i'll merge on a +1? want a sanity check real quick before i release and start tweeting :)

entrypoint.sh
```diff
@@ -6,43 +6,81 @@ export HOME="/github/workspace"
...
 
-if ! [ -z "$INPUT_WORKINGDIRECTORY" ]
+# If a working directory is detected as input
+if [ -n "$INPUT_WORKINGDIRECTORY" ]
 then
-  cd $INPUT_WORKINGDIRECTORY
+  cd "$INPUT_WORKINGDIRECTORY"
 fi
 
+# If an environmentdirectory is detected as input
```
EverlastingBugstopper:
```suggestion
# If an environment directory is detected as input
```

minddust:
switch to token auth was successful! thanks :) 🍰

Issue #16, opened 2020-02-01
Doesn't work for rust wasm worker


alank976:
It failed when it tried to install wasm-pack during `wrangler publish`
```
Wrangler has been installed!
+ @cloudflare/wrangler@1.6.0
added 27 packages from 13 contributors in 1.957s
⬇️ Installing wasm-pack...
 Compiling your project to WebAssembly...
Error: No such file or directory (os error 2)
Error: tried running command:
/github/workspace/.cache/.wrangler/wasm-pack-c6ddc20cc0d48874/wasm-pack build --target no-modules
```
I managed to reproduce it in my forked branch here https://github.com/alank976/wrangler-action/commit/2d7e8ad358457eea95676cad89f524edf213ab6e:
1. `wrangler generate test-rust-wasm https://github.com/cloudflare/rustwasm-worker-template` and tweak a bit `zone_id`, `route`, and `account_id` in wrangler.toml in order to suppress complaints about authentication
2. use docker-compose to reproduce the github action flow and mount the generated wasm project as `INPUT_WORKINGDIRECTORY`

You should be able to reproduce the same by running `docker-compose up`

I will probably try to fix by explicitly installing `wasm-pack` before `wrangler publish` and hopefully I can create a PR tomorrow for this :)

alank976:
After diving into it, basically this action doesn't support rust-wasm worker at the moment as its [`wasm-pack` requires Rust runtime](https://github.com/rustwasm/wasm-pack#-prerequisities).

There are a few solutions in my mind:
1. use different dockerfile conditionally based on `type` in `wrangler.toml`
1. use a dockerfile which has both node and rust installed
1. make a docker image available in dockerhub which is preinstalled wrangler, wasm-pack, and rust. Its version should be in sync with wrangler CLI version

signalnerve:
Ah, fascinating! We may want to revisit how we configure the Docker part of this to allow WASM deployments – you're totally right that this is JS-focused at the moment. cc @cloudflare/workers-devexp

ashleygwilliams:
wasm-pack is distributed as a prebuilt binary- so there's no "runtime" required... however wasm-pack does call the rust compiler to compile your rust to wasm, so the rust compiler does need to be installed.

happy to help figure this out and get out a new version!

alank976:
@signalnerve @ashleygwilliams Thank you for clarification. I am trying to make a PR with option 3 in my mind. Will tag you guys for code review :)

[...]

PR #18, opened 2020-02-06
Support rust wasm

alank976:
RE: issue #16:
This PR is about:
1. adding test-rust-wasm generated by wrangler CLI and rust wasm template
1. change the behavior of the docker image this action used by:
  - using rust as base, install nvm, nodejs and wrangler CLI during docker build (`RUN`)
  - tag/branch this github action as the version of installed wrangler
1. publish the all-in-one docker image to dockerhub (GPR is not useful as action user wont have right to pull the image easily)

That way, it supports rust-wasm, and save time for action users during run as most of the stuff are pre-installed.

I am totally new to wasm stuff. Please let me know your thoughts :)


xtuc:
I agree with @jjelev , I'm not comfortable with Cloudflare using an external image that we don't own.

alank976:
@jjelev @xtuc The idea is to illustrate the idea that it can be a pre-built docker image that has Rust, Node, and wrangler installed. It can be pulled from any cloudflare public docker registry (and it should be) if you guys agree with this approach. You can refer to the Dockerfile for how it is built and the GitHub action for how it is pushed.
So, apart from the implementation, what do you think in terms of the idea of having a pre-built, all-in-one docker which is versioned with that of the wrangler CLI?

PR #20, opened 2020-02-27
Remove nvm to speed build time

jasongill:
The Dockerfile already pulls the node:12 image, so there's no need for NVM to be downloaded and installed as well. Removing this speeds up the image build time (making Github Actions go faster).

From my testing, removing the un-needed nvm dependency cuts about 15-20 seconds off the Github Actions runtime, with no impact on functionality (as npm and node binaries are already included in the base Docker image, and those work just fine for wrangler).

From looking at the commit history, I think that this installation of NVM step is vestigial from before the node:12 image was used (and an Ubuntu image was being loaded instead).

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -3,19 +3,7 @@
 set -e
 
 export HOME="/github/workspace"
-export NVM_DIR="/github/workspace/nvm"
 export WRANGLER_HOME="/github/workspace"
-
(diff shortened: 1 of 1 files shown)


signalnerve:
Doing some testing on this today, so far it looks really good! Thanks :)

signalnerve:
This looks great, going to merge this!

signalnerve:
oops, did some fancy merging and this didn't seem to get closed, but it is merged into master. thanks so much!

PR #43, opened 2020-09-11
Update Readme to reference v1.3.0

chromy:
The readme mentions v1.2.0 but some of the commands (for example publish)
are only present in v1.3.0. Update readme to use v1.3.0 instead.

Thanks!


signalnerve:
ty

PR #56, opened 2021-05-13
Update Node.js to 16

ItalyPaleAle:
This makes it possible to build apps using Node.js 16, which has many new features (and will soon be the LTS version)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12
+FROM node:16
 ENV XDG_CONFIG_HOME /github/workspace
 ENV WRANGLER_HOME /github/workspace
 COPY entrypoint.sh /entrypoint.sh


olizilla:
This would also fix #59

jontybrook:
This would help me out too. Upvote!

jontybrook:
For the benefit of anyone else waiting for this to be merged, I have forked v1.3.0 and published a release "v2.0.0" (bumping major release as this is a major change) that runs on node 16. 

Whilst I will not be maintaining this fork, if you need to use node16 in your workflows, you can use our fork [carepenny/wrangler-action@2.0.0](https://github.com/carepenny/wrangler-action)

For example...

```     
...

     - name: Publish to Cloudflare
        uses: carepenny/wrangler-action@2.0.0
        with:
          apiToken: ${{ secrets.CF_API_TOKEN }}
```

felixoi:
This is also needed to support javascript modules

kaseyreed:
If folks are looking for a way to use a specific node version, you can do so using `actions/setup-node@v2` and specifying a no-op command for the `build` section in your `wrangler.toml`. 

For example:
```
# wrangler.toml
[build]
command = "echo 'do nothing bc you already built it'"

# your-action.yml
# ....
      - uses: actions/checkout@v2
      - uses: actions/setup-node@v2
        with:
          node-version: '16'
      - run: npm ci && npm run build
      - name: Publish
        uses: cloudflare/wrangler-action@1.3.0
        ...
```

eidam:
thank you for the contribution all, merging and we will create a new version soon

PR #58, opened 2021-06-11
Proposal: Allow specifying config

webbertakken:
**Context**
It can make sense to maintain two worker configurations in the same repository. 

For example:

- When using Flareact, or Nextjs with FABs for a frontend, your bundle is not exported as an ES Module.
- Subsequently when you want to use [durable objects](https://developers.cloudflare.com/workers/learning/using-durable-objects#instantiating-and-communicating-with-a-durable-object) an ES Module (.mjs) you're required to implement a default export that exposes an `async fetch` method (see last 2 paragraphs before the code). 

Combining both flavours in 1 file might not make sense, especially when prototyping.

The result is likely two config files for wrangler to use to publish the workers.

**Changes**

- This PR introduces a way to pass the path to a custom configFile, following the reference documentation from wrangler cli.
- Documentation for the new feature
- Feature is extendable through the new `$FLAGS` variable (flags represents both flags and options)
- Flags are now correctly provided before subcommand (see first screenshot)
- Fix a few typos 
- Fix missing fields in `action.yaml`

image

Proof that it works:

image


rvmourik:
Is there a specific reason this hasn't been merged yet? This would allow for a preview deployment whenever a PR is opened for example.

Or is Cloudflare working on integration preview deployment in workers as like the feature in Pages?

webbertakken:
I've been using this fork successfully for a while.

I'm also curious for a response :)

Scooletz:
Hey Cloudflare, it'd be great if you could merge it and publish. This would allow to use a single worker for multiple deployments with various `wrangler.toml` without the need of copying the code.

Scooletz:
@JacobMGEvans I know it's been 2 years, but it would be great if you could provide some reasoning behind closing this.

webbertakken:
I'm also still using this fork as there's no alternative it seems - apart from creating a monorepo with 2 modules and add the additional complexity on that side.

JacobMGEvans:
The whole thing is being rewritten, should be finished up next week. Would be open to considering it after the rewrite is done, feel free to open a feature ticket we can discuss further on 😊

webbertakken:
Excellent. Thank you!

PR #85, opened 2022-09-13
Update README.md

0xcaff:
Pass environment by input instead of variable. If you try to add a secret and use --env it puts the secret in the wrong place.


0xcaff:
Hmm seems like you need both actually nevermind

PR #87, opened 2022-09-27
mention environment input

jozan:
took me a while to figure out why my secrets did no update during deployment. now it's documented.


JacobMGEvans:

PR #100, opened 2023-03-06
Update to actions/checkout@v3 in README.md

Eusebiotrigo:
The first example still had the actions/checkout@v2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Deploy
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Publish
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Deploy
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
GrantBirki, line 24:
```suggestion
      - uses: actions/checkout@v3.4.0
```

It can be an even more up-to-date version as well

Eusebiotrigo: v3 should take the latest v3.x.x

GrantBirki: That is absolutely correct! Disregard my message in this case :)

Issue #84, opened 2022-09-12
🚀 Feature request: handle bulk upload of secrets from json


cameron-robey:
Now we have support in wrangler for bulk upload of secrets from a json file, it would be good to support this in `wrangler-action`.

See here for more information: https://github.com/cloudflare/wrangler2/pull/1799

alexdoesh:
I've submitted a PR, but if you need a quick solution that utilizes `secret:bulk` w/o creating/maintaining a new action module or waiting for PR to merge, you can try to add an extra step like this (I don't think it's the best possible solution, but it saves us ~30 seconds and doesn't look too insecure in our particular case):
```
      - name: Generate secrets JSON file
        run: |
          INPUT_SECRETS="SECRET1 SECRET2"
          TEMP_JSON_FILE="secrets.json"
          echo "{" > "$TEMP_JSON_FILE"
          first=true
          for SECRET in $INPUT_SECRETS; do
            VALUE=$(printenv "$SECRET") || secret_not_found "$SECRET"
            if [ "$first" = true ]; then
              first=false
            else
              echo "," >> "$TEMP_JSON_FILE"
            fi
            printf "\"%s\": \"%s\"" "$SECRET" "$VALUE" >> "$TEMP_JSON_FILE"
          done
          echo "}" >> "$TEMP_JSON_FILE"
        env:
          SECRET1: ${{ secrets.SECRET1 }}
          SECRET2: ${{ secrets.SECRET2 }}

      - name: Publish
        uses: cloudflare/wrangler-action@2.0.0
        with:
          apiToken: ${{ secrets.CF_API_TOKEN }}
          preCommands: |
            wrangler secret:bulk "/github/workspace/secrets.json"
            rm -f "/github/workspace/secrets.json"
```

PR #102, opened 2023-03-15
Add secret:bulk support

alexdoesh:
https://github.com/cloudflare/wrangler-action/issues/84


JacobMGEvans:

PR #108, opened 2023-05-24
Support wrangler3

### LINKED ISSUE #107: Update for wrangler v3
This repo needs updating to wrangler v3 officially. Namely the `publish` -> `deploy` command change should be the only major thing.

It might be a good opportunity to run through all the other pending issues and PRs in this repo too for some maintenance. 👀

nisshi-dev:
Upgrade to Wrangler v3 and related updates!

## Changes

- Upgraded Wrangler from v2 to v3
- Upgraded `azohra/shell-linter` from v0.3.0 to v0.6.0
- Upgraded `actions/checkout` from v2 to v3
- Upgraded Node.js from v16 to v18

## Issue
#107

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -32,9 +32,9 @@ secret_not_found() {
   exit 1
 }
 
-WRANGLER_VERSION=2
+WRANGLER_VERSION=3
 
-# If no Wrangler version is specified install v2.
+# If no Wrangler version is specified install v3.
 if [ -z "$INPUT_WRANGLERVERSION" ]; then
   npm i -g wrangler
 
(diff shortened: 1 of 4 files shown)


ShibaInu0413:
Change the instruction here from publish to deploy

https://github.com/cloudflare/wrangler-action/blob/a7c3c2e8a3b49fc1f4b1810e2a0ccd12e22510ef/entrypoint.sh#L141-L145

entrypoint.sh
```diff
@@ -43,16 +43,21 @@ elif [[ "$INPUT_WRANGLERVERSION" == 1* ]]; then
   npm i -g "@cloudflare/wrangler@$INPUT_WRANGLERVERSION"
   WRANGLER_VERSION=1
 
-# Else install Wrangler 2
+# If Wrangler version starts with 2 then install wrangler v2
+elif [[ "$INPUT_WRANGLERVERSION" == 2* ]]; then
+  npm i -g "@cloudflare/wrangler@$INPUT_WRANGLERVERSION"
```
WalshyDev, line 48: This block was only done for Wrangler 1 because the package changed. We do not need it (it also would not work since that's the wrangler 1 package, wrangler 2 is just `wrangler`)

entrypoint.sh
```diff
@@ -65,12 +70,12 @@ fi
...
 if [ -n "$INPUT_APIKEY" ] && [ -n "$INPUT_EMAIL" ]; then
 
-  # Wrangler v1 uses CF_ but v2 uses CLOUDFLARE_
+  # Wrangler v1 uses CF_ but v2 later uses CLOUDFLARE_
   if [ $WRANGLER_VERSION == 1 ]; then
     export CF_EMAIL="$INPUT_EMAIL"
     export CF_API_KEY="$INPUT_APIKEY"
   else
-    echo "::error::Wrangler v2 does not support using the API Key. You should instead use an API token."
+    echo "::error::Wrangler v2 later does not support using the API Key. You should instead use an API token."
```
WalshyDev, line 78: This is actually not correct English, it was ok before.

entrypoint.sh
```diff
@@ -134,9 +139,9 @@ if [ -z "$INPUT_COMMAND" ]; then
   echo "::notice:: No command was provided, defaulting to 'publish'"
 
  if [ -z "$INPUT_ENVIRONMENT" ]; then
-    wrangler publish
+    wrangler deploy
```
WalshyDev, line 142: We need to do this _only_ if it's v3+, right now this will break v1 and v2

WalshyDev (changes requested):

nisshi-dev:
Thank you everyone for your reviews. I have corrected the points. Please check.

entrypoint.sh
```diff
@@ -133,10 +138,18 @@ done
 if [ -z "$INPUT_COMMAND" ]; then
   echo "::notice:: No command was provided, defaulting to 'publish'"
 
- if [ -z "$INPUT_ENVIRONMENT" ]; then
-    wrangler publish
+  if [ -z "$INPUT_ENVIRONMENT" ]; then
+    if [ $WRANGLER_VERSION == 3 ]; then
```
StefanBratanov, line 142: Maybe `>=` is better to cover future version bumps.

JacobMGEvans:

PR #109, opened 2023-06-13
Support secret:bulk publish using file input

richardhunghhw:
Changes to add support for the [secret:bulk](https://developers.cloudflare.com/workers/wrangler/commands/#secretbulk) command as part of wrangler v2. 

Related issue - https://github.com/cloudflare/wrangler-action/issues/84

Tested by generating a JSON file from template using 1Password as the vault. Then uploading said file using the action.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ jobs:
         workingDirectory: 'subfoldername'
 ```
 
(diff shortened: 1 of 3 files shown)


JacobMGEvans:

PR #113, opened 2023-07-14
feat: add `quiet` parameter to silence stdout

winstxnhdw:
- Added `quiet` parameter to suppress certain stdout from Wrangler
- All process executions have this feature except `installWrangler`
- stdout in `uploadSecrets` is silenced by default as mentioned by @JacobMGEvans in #112


winstxnhdw:
Something's been up with both NPM and Yarn registries today. Might want to re-run the test workflow!

JacobMGEvans:
Going to consider this feature, for the current rewrite.

PR #132, opened 2023-08-08
Add Changesets

JacobMGEvans:
- The changeset action was added to the GitHub Actions workflow.
- This action will automatically create changesets for each pull request.
- Changesets are a way to track changes to a codebase.
- This will help to improve the development process by making it easier to track changes and understand the history of a codebase.


.github/workflows/changesets.yml
```diff
@@ -0,0 +1,35 @@
...
+          cache: "npm"
+
+      - name: Install modules
+        run: npm ci
+
+      - name: Create Version PR
+        id: changesets
+        uses: changesets/action@v1
+        with:
+          publish: npx changeset publish
```
1000hz, line 32: this will run `npm publish`, which we don't want. instead, I think we wanna do:
```suggestion
          createGithubReleases: true
```

.github/workflows/changesets.yml
```diff
@@ -0,0 +1,35 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - main
+
```
1000hz, line 7:
```suggestion

concurrency: ${{ github.workflow }}-${{ github.ref }}

```

.github/workflows/changesets.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: npm ci
+
+      - name: Create Version PR
+        id: changesets
+        uses: changesets/action@v1
+        with:
+          publish: npx changeset publish
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NODE_ENV: "production"
```
1000hz, line 35: why do we need to set `NODE_ENV`?

.changeset/stale-beers-promise.md
1000hz, line 1: we don't need to add a changeset / bump the version for this since it's not something that affects the package itself, just how it's released.

.github/workflows/changesets.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  release:
+    if: ${{ github.repository_owner == 'cloudflare' }}
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
1000hz, line 17: do we need this? this will cause all history for every branch and tag to be fetched instead of just the single latest commit.

.github/workflows/changesets.yml
JacobMGEvans: Naw, I can drop it

.changeset/stale-beers-promise.md
JacobMGEvans: I was wanting to use this PR to test it works.

.github/workflows/changesets.yml
JacobMGEvans: Mmm, probably not, I was doing that before for using SHA's and tags for various context in other workflows.

.github/workflows/changesets.yml
```diff
@@ -0,0 +1,34 @@
...
+    branches:
+      - main
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  release:
+    if: ${{ github.repository_owner == 'cloudflare' }}
+    name: Release
+    runs-on: ubuntu-latest
```
1000hz, line 14:
```suggestion
    runs-on: ubuntu-latest
    permissions:
      contents: write
      pull-requests: write
      issues: read
```

JacobMGEvans: You think this is needed, the Wrangler-SDK repo doesn't have them set up?

1000hz: workers-sdk grants all workflows read/write access at the repo level. this repo doesn't, and it would be better to explicitly scope permissions.

1000hz (approved):

Issue #137, opened 2023-08-09
v3 release currently broken


randomairborne:
[action + repo](https://github.com/randomairborne/speederboard/actions/runs/5813457248/job/15761064139) When trying to run the worker, iI get an "Error: Action Failed"

```
Run cloudflare/wrangler-action@v3
  with:
    apiToken: ***
    workingDirectory: cdn-upload-worker
    command: publish --env staging
📥 Installing Wrangler
  Running command: npm install wrangler@3.4.0
  npm WARN deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
  ✅ Wrangler installed
npm WARN deprecated rollup-plugin-inject@3.0.2: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-inject.
🚀 Executing Wrangler Commands
  Error: 🚨 Action failed
  ```

**Comment by AlCalzone** [comment]
Maybe because the action is using wrangler v3, which doesn't have a `publish` command anymore, but `deploy`?
https://developers.cloudflare.com/workers/wrangler/commands/#deploy

**Comment by JacobMGEvans** [comment]
I should add `stderr` for more commands.

**Comment by JacobMGEvans** [comment]
Also a heads up with the new `wrangler-action` you should be able to consolidate these (once I figure out your bug) 
```
      - name: deploy prod & dev
        uses: cloudflare/wrangler-action@v3
        with:
          apiToken: ${{ secrets.CF_API_TOKEN }}
          workingDirectory: "cdn-upload-worker"
          command:  | 
             deploy --config wrangler.prod.toml
             deploy --config wrangler.toml
```

**Comment by randomairborne** [comment]
ooh nice, thanks
let me know if you need anything else.

**Comment by randomairborne** [comment]
the issue was not publish

**Comment by JacobMGEvans** [comment]
> the issue was not publish

It will be, the release PR will be merged soon https://github.com/cloudflare/wrangler-action/pull/140

**Comment by randomairborne** [comment]
wait, publish is an issue?

**Comment by 1000hz** [comment]
No, @JacobMGEvans mistook your comment to mean "`cloudflare/wrangler-action` was not published" rather than
```diff
- command: publish --env staging
+ command: deploy --env staging
```
not solving your problem.

[...]

PR #141, opened 2023-08-10
Additional Error Handling

JacobMGEvans:
Previously, we prevented any error logs from propagating too far to prevent leaking of any potentially sensitive information. However, this made it difficult for developers to debug their code.

In this release, we have updated our error handling to allow for more error messaging from pre/post and custom commands. We still discourage the use of these commands for secrets or other sensitive information, but we believe this change will make it easier for developers to debug their code.

Example of output in `act` local run:
```ts
[...wrangler_action_self_testing]   ❓  ::group::🚀 Executing Wrangler Commands
[...wrangler_action_self_testing]   ❗  ::error::✘ [ERROR] Unknown arguments: dry-run, dryRun, asdf, deploy%0A%0A%0A
[...wrangler_action_self_testing]   ❓  ::endgroup::
```

Relates to #137

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -56,10 +56,28 @@ function semverCompare(version1: string, version2: string) {
 async function main() {
 	installWrangler();
 	authenticationSetup();
-	await execCommands(getMultilineInput("preCommands"), "Pre");
+	await execCommands(getMultilineInput("preCommands"), "pre");
 	await uploadSecrets();
 	await wranglerCommands();
-	await execCommands(getMultilineInput("postCommands"), "Post");
+	await execCommands(getMultilineInput("postCommands"), "post");
+}
+
+async function runProcess(
+	command: Parameters<typeof execAsync>[0],
+	options: Parameters<typeof execAsync>[1],
+) {
+	try {
+		const result = await execAsync(command, options);
+
+		result.stdout && info(result.stdout.toString());
+		result.stderr && error(result.stderr.toString());
+
+		return result;
+	} catch (err: any) {
+		err.stdout && info(err.stdout.toString());
+		err.stderr && error(err.stderr.toString());
+		throw err;
+	}
 }
 
 function checkWorkingDirectory(workingDirectory = ".") {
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -264,10 +268,16 @@ async function wranglerCommands() {
...
 		);
+
 		info(result.stdout);
 		return result;
 	});
-	await Promise.all(arrPromises);
+
+	await Promise.all(arrPromises).catch((result) => {
+		error(`🚨 Command failed`);
+		setFailed(result.stderr);
```
1000hz, line 278: While this is an improvement, this still will only output error messages if wrangler exits with a non-zero code (which it should, but doesn't always adhere to currently). I think it would be better to move the `result.stderr` logging to below the `info()` call in the block above (conditionally; we shouldn't log anything if `stderr` is empty). It's still good to call `setFailed()` at this level here, but with `🚨 Command failed` rather than the stderr message.

We also need to be sure to handle this for the `defaultCommandBehavior()`. Even better would be to consolidate these codepaths by instead setting a fallback value for `commands`.

src/index.ts
```diff
@@ -52,6 +53,10 @@ function semverCompare(version1: string, version2: string) {
 	return false;
 }
 
+function loggingResult(result: { stdout: string; stderr: string }) {
+	return result.stderr.length > 0 ? error(result.stderr) : info(result.stdout);
```
1000hz, line 57: I think we want to log both things, not just one or the other. My comment about making this conditional meant I don't want extra blank lines to be logged due to calling `error("")` when `result.stderr` is empty. We could do something like:
```ts
stdout && info(stdout)
stderr && error(stderr)
```

IMO, I wouldn't abstract just this logging logic into a separate function. It's not clear that this function performs a side-effect (which could be addressed with a different name), but more importantly the abstraction doesn't serve to simplify sufficiently complex logic to make the indirection worthwhile. it's roughly the same level of complexity, but with more ambiguity. it might make sense to wrap `execAsync`, however, to handle both executing a command and logging its result (including handling any resulting errors).

src/index.ts
```diff
@@ -264,10 +282,15 @@ async function wranglerCommands() {
...
 		);
-		info(result.stdout);
+
+		loggingResult(result);
 		return result;
 	});
-	await Promise.all(arrPromises);
+
+	await Promise.all(arrPromises).catch(() => {
+		setFailed(`🚨 Command failed`);
```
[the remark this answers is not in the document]
JacobMGEvans: lol

1000hz (approved):

PR #143, opened 2023-08-10
Update documentation to include new versioning syntax support

JacobMGEvans:


README.md
```diff
@@ -2,8 +2,12 @@
 
 Easy-to-use GitHub Action to use [Wrangler](https://developers.cloudflare.com/workers/cli-wrangler/). Makes deploying Workers, Pages or modifying R2 easy to do.
 
+## Big Changes in v3
+
 - Wrangler v1 is no longer supported.
 - Global API key & Email Auth no longer supported
+- Wrangler Action supports Action version syntax, i.e. `uses: cloudflare/wrangler-action@v3` or `uses: cloudflare/wrangler-action@v3.x` and will continue
+  to support `uses: cloudflare/wrangler-action@3.x.x` syntax.
```
lrapoport-cf, line 10:
```suggestion
- Action version syntax is newly supported. This means e.g. `uses: cloudflare/wrangler-action@v3` and `uses: cloudflare/wrangler-action@v3.x` are both now valid syntax. Previously supported syntax e.g. `uses: cloudflare/wrangler-action@3.x.x` continues to be supported.
```

what do you think of this suggestion? i'm trying to keep the style (i.e. passive voice) consistent with the other bullets, but let me know if this changes the meaning in a way that's not intended

lrapoport-cf (approved):

PR #161, opened 2023-08-29
Refactor error handling to stop execution on errors

### LINKED ISSUE #160: [Bug] Secrets upload fails, still uploads Worker
If the Secrets upload fails it should fail fast and now allow the Worker to deploy.

1000hz:
Fixes #160

This PR refactors the way errors are propagated through the action when encountered. Previously, the action relied solely on `core.setFailed()` to handle errors, but this only sets the action status while allowing execution to continue. You can see this occur in https://github.com/cloudflare/wrangler-action/actions/runs/5968662403/job/16192935017#step:9:22, where the secret upload step failed, but the action proceeded to deploy the worker anyway.

Now instead of calling `core.setFailed()` at the source of every failure case, we throw errors and catch them at the top level where we then output the error and call `core.setFailed()`.

This PR also:
* tweaks a few error messages to be more descriptive.
* removes a couple of places where the same errors were being handled at multiple levels.
* adds `finally` blocks around some `endGroup()` calls to ensure no groups are left dangling open if an error occurs.

### DIFF
--- a/.changeset/stale-spiders-turn.md
+++ b/.changeset/stale-spiders-turn.md
@@ -1,0 +1,5 @@
+---
+"wrangler-action": patch
+---
+
+Refactored error handling to stop execution when action fails. Previously, the action would continue executing to completion if one of the steps encountered an error. Fixes #160.
--- a/src/index.ts
+++ b/src/index.ts
@@ -80,13 +80,18 @@ function semverCompare(version1: string, version2: string) {
 }
 
 async function main() {
-	installWrangler();
-	authenticationSetup();
-	await execCommands(getMultilineInput("preCommands"), "pre");
-	await uploadSecrets();
-	await wranglerCommands();
-	await execCommands(getMultilineInput("postCommands"), "post");
-	info("🏁 Wrangler Action completed", true);
+	try {
+		installWrangler();
+		authenticationSetup();
+		await execCommands(getMultilineInput("preCommands"), "pre");
+		await uploadSecrets();
+		await wranglerCommands();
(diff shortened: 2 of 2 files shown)


src/utils.ts
JacobMGEvans, line 1: This seems like a premature abstraction, the file is far from being large and colocating the functions means not having to jump around files.

JacobMGEvans (approved):

tsconfig.json
```diff
@@ -9,12 +9,12 @@
 		"isolatedModules": true,
 		"target": "ESNext",
 		"module": "ESNext",
-		"moduleResolution": "NodeNext",
+		"moduleResolution": "Bundler",
```
JacobMGEvans, line 12: Oh that is cool! I didnt see they had a new [moduleResolution `bundler`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler)

JacobMGEvans:
Please add a changeset for this too! 😄

src/utils.ts
1000hz: Importing anything from `index.ts` has the side-effect of executing the action since `main()` is executed in top-level scope. This meant that running the vitest test suite to test these util functions was inadvertently executing the entire action. Pulling these into their own file felt like the least disruptive way around that, especially since they’re already being unit tested in isolation. Alternatively I could have renamed `index.ts` to `main.ts` and added a new `index.ts` that just imports and executes `main()`, but that would have made this PR harder to review and I didn’t want to make the diff any noisier than it’s already become.

PR #165, opened 2023-08-31
Build action before pushing dist-tags

1000hz:
Another followup to #159. We forgot to move the build step over.


JacobMGEvans:
I legit thought it just did the build for you for some reason.

JacobMGEvans (approved):

PR #171, opened 2023-09-15
Refactor subprocess execution to use `@actions/exec`

### LINKED ISSUE #168: Secrets truncation
Uploading a worker with many secrets only partially uploads the secrets but no error or warning is emitted - all success. Looking in the worker dashboard, the secret is simply not listed.
We're way under the 128 combined vars/secrets limit.
Started noticing this behavior on the 11th secret. And the extra secrets (past 10) that doesn't get updated appears to be random (only tested on 3 builds - need more data to be sure about this.
Actions is set to use `latest` worker version
```yaml
- uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d # v3.8.1
  with:
    node-version: lts/*
    check-latest: true
    cache: 'npm'
- uses: cloudflare/wrangler-action@09fc6b504568cf179b085df24393f237fea6c7f6 # v3.1.1
  with:
    wranglerVersion: latest
    apiToken: ${{ secrets... }}
    accountId: ${{ vars... }}
    vars: |
      ...
      ...
    secrets: |
      ...
      ...
      ...
      ...
      ...
      ...
      ...
      ...
      ...
      ...
      ...
  env:
    ...: ${{ vars...}}
    ...: ${{ github...}}
    ...: ${{ secrets...}}
    ...: ${{ secrets...}}
    ...: ${{ secrets...}}
    ...: ${{ secrets... }}
    ...: ${{ secrets... }}
    ...: ${{ secrets... }}
    ...: ${{ secrets...}}
    ...: ${{ secrets... }}
    ...: ${{ secrets...}}
    ...: ${{ secrets... }}
    ...: ${{ secrets...}}
```
image
This is not a new worker (so I don't think its related to #162), but these are new secrets (compared to previous deployments of the same worker)

1000hz:
Instead of using a mix of `child_process.exec`, `child_process.execSync` and a promisified version of `child_process.exec`, we now (mostly) just use [`@actions/exec`](https://github.com/actions/toolkit/blob/main/packages/exec/README.md). That runs `child_process.spawn` under the hood and handles a lot of character escaping for us. We can also now pass Buffers directly into the subprocess as stdin instead of relying on shell piping.

This ends up fixing a few problems we had where secrets and env var values containing shell metacharacters were being misinterpreted.

Unfortunately, `@actions/exec` doesn't support running with a shell. That means we still have to roll our own wrapper around `child_process.exec` to avoid a breaking change to `preCommands` and `postCommands`, since users might be expecting these to run within a shell.

Also worth noting that we're no longer hiding stdout and stderr during the secret uploading step. We were previously doing this out of an abundance of caution, but it made debugging issues very difficult if secret upload failed for some reason. I feel ok doing this since we're no longer echoing & piping the secret values, wrangler doesn't ever output secret values, and as a last line of defense GitHub masks any secret values that accidentally get logged.

Fixes #168

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,11 +8,9 @@ import {
 	startGroup as originalStartGroup,
 	setFailed,
 } from "@actions/core";
-import { exec, execSync } from "node:child_process";
-import * as util from "node:util";
+import { exec, execShell } from "./exec";
 import { checkWorkingDirectory, semverCompare } from "./utils";
 import { getPackageManager } from "./packageManagers";
-const execAsync = util.promisify(exec);
 
 const DEFAULT_WRANGLER_VERSION = "3.5.1";
 
@@ -62,8 +60,8 @@ function endGroup(): void {
 
 async function main() {
 	try {
-		installWrangler();
 		authenticationSetup();
+		await installWrangler();
 		await execCommands(getMultilineInput("preCommands"), "pre");
 		await uploadSecrets();
 		await wranglerCommands();
@@ -75,36 +73,28 @@ async function main() {
 	}
 }
 
-async function runProcess(
-	command: Parameters<typeof execAsync>[0],
-	options: Parameters<typeof execAsync>[1],
-) {
-	try {
-		const result = await execAsync(command, options);
-
-		result.stdout && info(result.stdout.toString());
-		result.stderr && error(result.stderr.toString(), true);
-
-		return result;
-	} catch (err: any) {
-		err.stdout && info(err.stdout.toString());
-		err.stderr && error(err.stderr.toString(), true);
-		throw new Error(`\`${command}\` returned non-zero exit code.`);
-	}
-}
-
-function installWrangler() {
+async function installWrangler() {
 	if (config["WRANGLER_VERSION"].startsWith("1")) {
 		throw new Error(
 			`Wrangler v1 is no longer supported by this action. Please use major version 2 or greater`,
 		);
 	}
+
 	startGroup("📥 Installing Wrangler");
-	const command = `${packageManager.install} wrangler@${config["WRANGLER_VERSION"]}`;
-	info(`Running command: ${command}`);
-	execSync(command, { cwd: config["workingDirectory"], env: process.env });
-	info(`✅ Wrangler installed`, true);
-	endGroup();
+	try {
+		await exec(
+			packageManager.install,
+			[`wrangler@${config["WRANGLER_VERSION"]}`],
+			{
+				cwd: config["workingDirectory"],
+				silent: config["QUIET_MODE"],
+			},
+		);
+
+		info(`✅ Wrangler installed`, true);
+	} finally {
+		endGroup();
+	}
 }
 
 function authenticationSetup() {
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -113,62 +99,65 @@ async function execCommands(commands: string[], cmdType: string) {
 
 	startGroup(`🚀 Running ${cmdType}Commands`);
 	try {
-		const arrPromises = commands.map(async (command) => {
+		for (const command of commands) {
```
RamIdeas, line 102: These commands were running in parallel, now sequentially. Is that intentional? Is that a breaking change?

src/index.ts
```diff
@@ -113,62 +99,65 @@ async function execCommands(commands: string[], cmdType: string) {
...
-		);
-
-	await Promise.all(arrPromises);
+	return Promise.all(
+		secrets.map((secret) => {
+			const args = ["wranger", "secret", "put", secret];
+			if (environment) {
+				args.push("--env", environment);
+			}
+			return exec(getNpxCmd(), ["wrangler", "secret", "put", secret], {
```
RamIdeas, line 154: `args` isn't being used so `--env` isn't ever applied

src/index.ts
```diff
@@ -242,27 +214,29 @@ async function wranglerCommands() {
 			commands.push(deployCommand);
 		}
 
-		const arrPromises = commands.map(async (command) => {
-			if (environment.length > 0 && !command.includes(`--env`)) {
-				command = command.concat(` --env ${environment}`);
-			}
+		for (let command of commands) {
+			const args = ["wrangler", ...command.split(/\s+/)];
```
RamIdeas, line 218: This is hoping the command has no quotes strings with spaces

We might want to document this to also allow `command` to be an array as an escape hatch

RamIdeas (changes requested):

src/index.ts
1000hz: Yes, this is intentional, though I forgot to include a changeset for it. The previous behavior ought to be considered a bug, since ostensibly no one expects these to run in parallel and it could introduce race conditions. I don’t think this is a breaking change, but it would be had we gone from sequential to parallel execution.

src/index.ts
1000hz: Oops, good catch!

src/index.ts
1000hz: I think `@actions/exec` handles that scenario for us with its escaping logic, but worth looking into to confirm that.

JacobMGEvans (commented):
Just checking on this?

src/index.ts
1000hz: you were right, great catch. I've removed the `split` here and instead now pass the provided command (prepended with `npx wrangler `) as part of the first argument to `exec()`, which is expected to be a string that has already been escaped. The second array argument is now used to provide any additional options we want to inject into the users command (e.g. `--env ...` or `--vars ...`), which `exec()` will escape for us.

src/index.ts
1000hz: fixed

RamIdeas (approved):

Issue #181, opened 2023-10-05
Error Installing Wrangler with pnpm


saviski:
This error started happening with the lates version

```
📥 Installing Wrangler
  Running command: pnpm add wrangler@3.5.1
  Error: Command failed: pnpm add wrangler@3.5.1
  Error: 🚨 Action failed
```

1000hz:
It seems like this could be a result of `pnpm` not being available in the runner environment. Are you including https://github.com/pnpm/action-setup in an earlier step of your workflow before `wrangler-action`?

AdiRishi:
I just ran into this myself. For me I think the special case was that I'm using the `workingDirectory` property

```yml
- name: Deploy apex-gateway
  uses: cloudflare/wrangler-action@v3
  with:
    accountId: ${{ secrets.CF_ACCOUNT_ID || secrets.CLOUDFLARE_ACCOUNT_ID }}
    apiToken: ${{ secrets.CF_API_TOKEN || secrets.CLOUDFLARE_API_TOKEN }}
    workingDirectory: 'workers/apex-gateway'
    command: 'deploy'
```

My suspicion is that it is looking for `pnpm-lock.yaml` which doesn't exist when running in the subdirectory.

AdiRishi:
Actually I'm now 100% sure that it's a bug because it's searching in the workingDirectory. We can see this in the codebase
```ts
function detectPackageManager(
	workingDirectory = ".",
): PackageManagerValue | null {
	if (existsSync(path.join(workingDirectory, "package-lock.json"))) {
		return "npm";
	}
	if (existsSync(path.join(workingDirectory, "yarn.lock"))) {
		return "yarn";
	}
	if (existsSync(path.join(workingDirectory, "pnpm-lock.yaml"))) {
		return "pnpm";
	}
	if (existsSync(path.join(workingDirectory, "bun.lockb"))) {
		return "bun";
	}
	return null;
}
```

This is likely a seperate issue to the one being reported, should I create a new issue report?

1000hz:
@AdiRishi yes, please open a new issue for that

AdiRishi:
> @AdiRishi yes, please open a new issue for that

Done - https://github.com/cloudflare/wrangler-action/issues/198

thantos:
> It seems like this could be a result of `pnpm` not being available in the runner environment. Are you including https://github.com/pnpm/action-setup in an earlier step of your workflow before `wrangler-action`?

[...]

PR #207, opened 2023-11-05
fix: detect pnpm-workspace.yml and append -w when installing wrangler

sam-goodwin:
The wrangler-action does not currently support pnpm workspaces. 

See: https://github.com/cloudflare/wrangler-action/issues/181#issuecomment-1765527537

It fails with:
```
📥 Installing Wrangler
  /home/runner/setup-pnpm/node_modules/.bin/pnpm add wrangler@* -w
   ERR_PNPM_ADDING_TO_ROOT  Running this command will add the dependency to the workspace root, which might not be what you want - if you really meant it, make it explicit by running this command again with the -w flag (or --workspace-root). If you don't want to see this warning anymore, you may set the ignore-workspace-root-check setting to true.
Error: The process '/home/runner/setup-pnpm/node_modules/.bin/pnpm' failed with exit code 1
Error: 🚨 Action failed
```

This change detects a pnpm workspace by also looking for `pnpm-workspace.yml` file and then selects an install script that appends `-w`

### DIFF
--- a/src/packageManagers.ts
+++ b/src/packageManagers.ts
@@ -19,6 +19,10 @@ const PACKAGE_MANAGERS = {
 		install: "pnpm add",
 		exec: "pnpm exec",
 	},
+	"pnpm-workspace": {
+		install: "pnpm add -w",
+		exec: "pnpm exec",
+	},
 	bun: {
 		install: "bun i",
 		exec: "bunx",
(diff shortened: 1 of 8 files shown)


src/packageManagers.ts
```diff
@@ -19,6 +19,10 @@ const PACKAGE_MANAGERS = {
 		install: "pnpm add",
 		exec: "pnpm exec",
 	},
+	"pnpm-workspace": {
+		install: "pnpm add -w",
+		exec: "pnpm exec",
```
sam-goodwin, line 24: I don't think `pnpm exec` needs to be adjusted.

sam-goodwin:
I noticed the tests only check that the right config is selected. It would be better to actually run the action within the test folders instead.

lrapoport-cf:
hey @sam-goodwin :) we haven't heard from you in a while. if you'd still like to get this PR landed, please update in response to @1000hz 's feedback. otherwise, we'll close this PR until you're ready to revisit :) thanks!

lrapoport-cf:
hey @sam-goodwin :) we haven't heard from you so we'll close this PR for now, but please feel free to re-open when you're ready to revisit. thanks!

jacob-shuman:
This is currently preventing me from using the wrangler action in my pnpm monorepo. I think this can be automatically inferred (at least for pnpm) by checking for a `pnpm-workspace.yaml`/`pnpm-workspace.yml` file in the cwd.

Issue #211, opened 2023-11-22
Add outputs to action


1000hz:
As requested in #210, it would be useful for this action to provide some [outputs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#outputs-for-docker-container-and-javascript-actions) to allow users to consume the results of the action run in subsequent workflow steps.

For the MVP, we could start with providing a `command-output` that makes the stdout from executing `command` available. We could also possibly try to detect if a workers or pages deployment was made and provide a `deployment-url` we infer from the output.

admah:
Would solving [this issue](https://github.com/cloudflare/workers-sdk/issues/2012) help with this?

Issue #210, opened 2023-11-21
Commands piped to Wrangler doesn't work in GitHub Actions


vipulgupta2048:
Here's our GitHub action step using Wrangler Action.

```
    - name: Deploy to Cloudflare Pages
        uses: cloudflare/wrangler-action@5e8484995321734668f14981c316aa9188d76ed1 # v3.3.2
        with:
          apiToken: ${{ secrets.CF_API_TOKEN }}
          accountId: ${{ secrets.CF_ACCOUNT_ID }}
          wranglerVersion: "3.5.1" 
          preCommands: wrangler --version
          postCommands: wrangler pages deploy --branch ${{ env.CF_BRANCH }} --project-name=${{ inputs.cloudflare_website }} build/ | tee -a $GITHUB_STEP_SUMMARY
```

This command used to work, but is now failing when trying out the new wrangler-action v3.3.2 with the following error. We used the pipe to get command output into GitHub Step Summary for a nice bit of result and get the Cloudflare pages draft URL.  

Here's the error: The wrangle command tries to check what these arguments are as well. 
```
🚀 Running Wrangler Commands
  /home/runner/_work/_tool/node/18.18.2/x64/bin/npx wrangler pages deploy --branch renovate/cloudflare-wrangler-action-3.3.x --project-name=flowzone build/ | tee -a $GITHUB_STEP_SUMMARY`


  ✘ [ERROR] Unknown arguments: a, |, tee
```

I recommend that we define proper outputs for this action in the `action.yml` file. These standardized outputs can be used by folks anywhere? Can add that myself if people agree?

Link of the action run: https://github.com/product-os/flowzone/actions/runs/6905675937/job/18789164936

vipulgupta2048:
Heyyo @1000hz From my testing, I understood as much about `preCommands/postCommands`. I was thinking about this suggestion that you made as the only way out but wanted to confirm if the removal of piping in `command` is expected. Will make the change as you suggested, thanks!

1000hz:
Yes, removing support for shell piping in `command` was intentional since it probably should have never been allowed given how we treat the input. In hindsight, we should have done a major version bump when we made this change because it effectively broke anyone relying on this behavior.

vipulgupta2048:
Yeah no worries @1000hz These things happen. Happy to help improve this bit.

PR #213, opened 2023-12-01
Save Wrangler Command Output to a Variable

GrantBirki:
# Save Wrangler Command Output to a Variable

> Please disregard the branch name, I changed what this PR does per @1000hz's suggestions

This pull request enables a more advanced usage of this Action in conjunction with other workflow steps. These changes allow users to gain the ability to parse the wrangler command's output in subsequent workflow steps should they need to.

The [`@actions/exec`](https://github.com/actions/toolkit/tree/main/packages/exec) package is utilized to save the `stdout` and `stderr` streams to variables and then those variables are written to two new Actions outputs when this job completes. These two new outputs are as follows:

- `command-output`: contains the string results of `stdout`
- `command-stderr`: contains the string results of `stderr`

But wait, there is more! ⭐ 

Since we are saving the `stdout` to a variable, it can now be parsed by this Action just a *little bit more* before exiting. I added in a bonus feature to this PR to detect if the command being run is a deployment for a Cloudflare Pages site or Worker. If it is a deployment, then the `deployment-url` output gets populated.

- `deployment-url`: contains the string results of the URL that was deploy (ex: `https://<your_pages_site>.pages.dev`)

## Why?

I personally have several open source projects which I help maintain. We use this Action to deploy our Cloudflare Pages sites. Recently, our workflows for `dev` environments started breaking. This was a direct result of https://github.com/cloudflare/wrangler-action/issues/210

> Here is a resulting Action's log of the [failure](https://github.com/the-hideout/tarkov-dev/actions/runs/7061169488/job/19227893794#step:7:37)

This is because we are trying to save the `stdout` of this Action by piping it to a file. This isn't really the best solution but it _worked_... at least for a while before it broke. This PR will prevent us from having to use choppy command line hacks in the future.

Being able to capture the output of this command to get the unique `dev` environment URL for pages deployments is required by our workflows to use it in subsequent workflow steps. This pull request also solves that problem via the new `deployment-url` output that gets saved as well.

## Example :camera_flash:

image

> live example workflow -> https://github.com/the-hideout/tarkov-dev/actions/runs/7064233377/job/19231836343

---

resolves: https://github.com/cloudflare/wrangler-action/issues/211

Thanks! :star:

### DIFF
--- a/.changeset/plenty-seals-cover.md
+++ b/.changeset/plenty-seals-cover.md
@@ -1,0 +1,11 @@
+---
+"wrangler-action": minor
+---
+
+This change introduces three new GitHub Action's output variables. These variables are as follows:
+
+- `command-output` - contains the string results of `stdout`
+- `command-stderr` - contains the string results of `stderr`
+- `deployment-url` - contains the string results of the URL that was deployed (ex: `https://<your_pages_site>.pages.dev`)
+
+These output variables are intended to be used by more advanced workflows that require the output results or deployment url from Wrangler commands in subsequent workflow steps.
--- a/.changeset/plenty-seals-cover.md
+++ b/.changeset/plenty-seals-cover.md
@@ -2,7 +2,7 @@
 "wrangler-action": minor
 ---
 
-This change introduces three new GitHub Action's output variables. These variables are as follows:
+This change introduces three new GitHub Actions output variables. These variables are as follows:
 
 - `command-output` - contains the string results of `stdout`
 - `command-stderr` - contains the string results of `stderr`
--- a/README.md
+++ b/README.md
@@ -218,6 +218,37 @@ jobs:
 
 For more advanced usage or to programmatically trigger the workflow from scripts, refer to [the GitHub documentation](https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event) for making API calls.
 
+## Advanced Usage
+
+### Saving Wrangler Command Output to a File
+
+More advanced workflows may need to parse the resulting output of Wrangler commands. To do this, you can use the `outputFile` option to save the output of the command to a file. This file will be available to subsequent steps in the workflow. The format of this file is JSON so that it can be easily parsed by other steps.
+
+```yaml
+- name: Deploy
+  id: deploy
+  uses: cloudflare/wrangler-action@v3
+  with:
+    apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+    accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+    command: pages deploy --project-name=example
+    outputFile: "true"
+
+- name: print wrangler command output
+  env:
+    OUTPUT_PATH: ${{ steps.deploy.outputs.wranglerCommandOutputFile }}
+  run: cat $OUTPUT_PATH
+```
+
+Now when you run your workflow, you will see the output of the Wrangler command in the logs that should be in the following structure:
+  
+  ```json
+  {
+    "stdout": "stdout text here",
+    "stderr": "stderr text here"
+  }
+  ```
+
 ## Troubleshooting
 
 ### "I just started using Workers/Wrangler and I don't know what this is!"
--- a/README.md
+++ b/README.md
@@ -239,14 +239,7 @@ More advanced workflows may need to parse the resulting output of Wrangler comma
   run: echo $CMD_OUTPUT
 ```
 
-Now when you run your workflow, you will see the output of the Wrangler command in the logs that should be in the following structure:
-  
-  ```json
-  {
-    "stdout": "stdout text here",
-    "stderr": "stderr text here"
-  }
-  ```
+Now when you run your workflow, you will see the full output of the Wrangler command in your workflow logs. You can also use this output in subsequent workflow steps to parse the output for specific values.
 
 ## Troubleshooting
 
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ Now when you run your workflow, you will see the full output of the Wrangler com
 
 ### Using the `deployment-url` Output Variable
 
-If you are executing a Wrangler command that results in either a workers or pages deployment, you can utilize the `deployment-url` output variable to get the URL of the deployment. For example, if you want to print the deployment URL after deploying your application, you can do the following:
+If you are executing a Wrangler command that results in either a Workers or Pages deployment, you can utilize the `deployment-url` output variable to get the URL of the deployment. For example, if you want to print the deployment URL after deploying your application, you can do the following:
 
 ```yaml
 - name: Deploy
--- a/README.md
+++ b/README.md
@@ -241,6 +241,8 @@ More advanced workflows may need to parse the resulting output of Wrangler comma
 
 Now when you run your workflow, you will see the full output of the Wrangler command in your workflow logs. You can also use this output in subsequent workflow steps to parse the output for specific values.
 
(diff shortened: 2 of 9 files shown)


Razzmatazzz (approved):

Shebuka (approved):

1000hz:
Hey @GrantBirki, thanks for the PR. Would your use case be satisfied if we implemented the `command-output` action output as described in #211?

GrantBirki:
@1000hz that would be even better! Do you mind if I alter this PR to make all the text that gets written to a file be written to the `command-output` output instead?

package.json
1000hz, line 1: This action manages versioning via Changesets. Please run `npx changeset` to add a new changeset file describing the user-facing changes instead of updating the version directly.

src/index.ts
```diff
@@ -242,10 +243,45 @@ async function wranglerCommands() {
...
+
+			// Set the outputs for the command
+			setOutput("command-output", stdOut);
+			setOutput("command-stderr", stdErr);
+
+			// Check if this command is a workers or pages deployment
+			if (
+				command.startsWith("deploy") ||
+				command.startsWith("publish") ||
+				command.startsWith("pages publish")
```
1000hz, line 275: We also need to check for `pages deploy`

src/index.ts
```diff
@@ -242,10 +243,45 @@ async function wranglerCommands() {
...
+			) {
+				// If this is a workers or pages deployment, try to extract the deployment URL
+				let deploymentUrl = "";
+				const deploymentUrlMatch = stdOut.match(/https?:\/\/[a-zA-Z0-9-\.\/]+/);
```
1000hz, line 279: while what you have still works, it's not necessary to escape the `.` and `/` characters inside of character class expressions.

src/index.ts
```diff
@@ -242,10 +243,45 @@ async function wranglerCommands() {
...
+					},
+				},
+			};
+
+			// Execute the wrangler command
+			await exec(`${packageManager.exec} wrangler ${command}`, args, options);
+
+			// Set the outputs for the command
+			setOutput("command-output", stdOut);
+			setOutput("command-stderr", stdErr);
```
1000hz, line 269: the `command` input accepts multiline strings and will execute each line as a separate command execution. as currently implemented, `command-output` and `command-stderr` will only capture output from the last executed command within the loop. this is _probably ok_ for now, but I wonder if it would be more useful to collect (concatenated) output from all command executions, or if that would be confusing without some delimiter between execution results.

action.yml
```diff
@@ -44,3 +44,10 @@ inputs:
   packageManager:
     description: "The package manager you'd like to use to install and run wrangler. If not specified, the preferred package manager will be inferred based on the presence of a lockfile or fallback to using npm if no lockfile is found. Valid values are `npm` | `pnpm` | `yarn` | `bun`."
     required: false
+outputs:
+  command-output:
+    description: "The output of the Wrangler command (comes from stdout)"
+  command-stderr:
+    description: "The error output of the Wrangler command (comes from stderr)"
+  deployment-url:
+    description: "If the command was a workers or pages deployment, this will be the URL of the deployment"
```
1000hz, line 53:
```suggestion
    description: "If the command was a Workers or Pages deployment, this will be the URL of the deployment"
```

1000hz (changes requested):
Nice work, @GrantBirki! Left a few comments.

src/index.ts
GrantBirki: I am personally and end-user of the exact changes that I am implementing and I would personally find the extra output of other commands in these variables to be confusing. If we added concatenated output in some form, (via a delimiter) this would likely lead to even more string formatting / processes that I was trying to ultimately get farther away from with these changes. For now, I think this is okay and would be willing to iterate on a future pull request if folks think that would be a good idea or have different use cases in the future.

src/index.ts
GrantBirki, line 275:
```suggestion
				command.startsWith("pages publish") ||
				command.startsWith("pages deploy")
```

src/index.ts
GrantBirki: Perhaps I'm not familiar enough with regular expressions in JavaScript but I would think that removing the escape characters would cause all sorts of errors here. 

I got a fair amount of errors when trying to test this out so perhaps we should leave it the way that it is for now since it is working as expected?

Reference:

image

src/index.ts
GrantBirki: Added!

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "wrangler-action",
-	"version": "3.3.2",
+	"version": "3.4.0",
```
GrantBirki, line 3:
```suggestion
	"version": "3.3.2",
```

I should not have edited the version directly here according to https://github.com/cloudflare/wrangler-action/pull/213#discussion_r1424532205

package.json
GrantBirki: I reverted the version changes that I made manually via https://github.com/cloudflare/wrangler-action/pull/213/commits/9dd38ad7904f65b85f8c11fc181be64f854f9f25 and will add a new changeset via `npx changeset` tomorrow 🙇

src/index.ts
1000hz: Sorry for the confusion. To be clear, I’m referring only to the `\.\/` _within the character class expression_ (the stuff between `[]`), not the `\/\/` outside of it, which do still need to be escaped. 
```suggestion
				const deploymentUrlMatch = stdOut.match(/https?:\/\/[a-zA-Z0-9-./]+/);
```

src/index.ts
1000hz: I’m ok with that.

package-lock.json
1000hz, line 1: We also want to remove this change to the lockfile

src/index.ts
GrantBirki: Oh interesting! I went ahead and committed your suggestion

package.json
GrantBirki: Done!

.changeset/plenty-seals-cover.md
```diff
@@ -0,0 +1,11 @@
+---
+"wrangler-action": minor
+---
+
+This change introduces three new GitHub Action's output variables. These variables are as follows:
```
1000hz, line 5:
```suggestion
This change introduces three new GitHub Actions output variables. These variables are as follows:
```

README.md
```diff
@@ -218,6 +218,56 @@ jobs:
...
+  run: echo $CMD_OUTPUT
+```
+
+Now when you run your workflow, you will see the full output of the Wrangler command in your workflow logs. You can also use this output in subsequent workflow steps to parse the output for specific values.
+
+> Note: the `command-stderr` output variable is also available if you need to parse the standard error output of the Wrangler command.
+
+### Using the `deployment-url` Output Variable
+
+If you are executing a Wrangler command that results in either a workers or pages deployment, you can utilize the `deployment-url` output variable to get the URL of the deployment. For example, if you want to print the deployment URL after deploying your application, you can do the following:
```
1000hz, line 248:
```suggestion
If you are executing a Wrangler command that results in either a Workers or Pages deployment, you can utilize the `deployment-url` output variable to get the URL of the deployment. For example, if you want to print the deployment URL after deploying your application, you can do the following:
```

1000hz (approved):

1000hz:
Thanks for the contribution, @GrantBirki!

GrantBirki:
Thank you! :tada:

PR #216, opened 2023-12-09
fix: semver comparison

### LINKED ISSUE #215: action doesn't use bulk secret upload when specifying wranglerVersion 3.10.0 or higher
There's a bug in our `semverCompare()` function resulting from comparing version numbers lexicographically. This results in places where we use it (e.g. whether to use bulk secret uploads or serial uploads) sometimes taking the wrong code path.

Cherry:
fixes #215 
closes #214 

This also bumps a few dependencies, to remove some `npm audit` warnings. The output bundle increases from `964kB` to `983kB`, but I feel the 19kB increase is worth the reliable semver comparison.

### DIFF
--- a/.changeset/weak-lobsters-march.md
+++ b/.changeset/weak-lobsters-march.md
@@ -1,0 +1,9 @@
+---
+"wrangler-action": patch
+---
+
+Fixes issues with semver comparison, where version parts were treated lexicographically instead of numerically.
+
+Bulk secret uploading was introduced in wrangler `3.4.0`, and this action tries to check if the version used is greater than `3.4.0`, and then if so, using the new bulk secret API which is faster. Due to a bug in the semver comparison, `3.19.0` was being considered less than `3.4.0`, and then using an older and slower method for uploading secrets.
+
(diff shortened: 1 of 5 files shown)


petebacondarwin (approved):

Cherry:
Not sure what's up with the test failure here, that's not one that I touched 🤔

petebacondarwin:
The failed test is due to updating vitest to v1, which contains this breaking change: https://github.com/vitest-dev/vitest/pull/4396.

The resolution is just to fix up the snapshot to match the new expected value.

Cherry:
Great catch, thanks @petebacondarwin. I've updated that now and all should be good.

petebacondarwin:
Hmm. The failing tests need to be run on our fork since they require secrets...

Cherry:
Yeah I can't really fix that one unfortunately 😞 I've not made any real changes to the deploy process of this, but let me know if I can assist with anything further here to get this merged.

petebacondarwin:
Well that is interesting! By pushing a branch with the exact commit from this PR in it, the CI tests ran and Github attributed the successful test run to this PR too! TIL!

All good to go.

Thanks @Cherry

PR #227, opened 2024-01-03
(enhancement): Surface inner exception when secret:bulk upload command fails

### LINKED ISSUE #226: (enhancement): Surface inner exception when uploading secrets fails
There has been some scattered discussion on this in other issues but I'm making a seperate issue here as I plan on making a PR for this.

Currently, whenever there is an error uploading secrets, the `wrangler-action` shows `Failed to upload secrets.` but provides no additional context.
Screenshot 2024-01-02 at 2 56 08 pm

This makes it very very difficult to debug and have any idea on why the secret upload failed.

AdiRishi:
Resolves #226

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -208,7 +208,7 @@ async function uploadSecrets() {
 	} catch (err: unknown) {
 		if (err instanceof Error) {
 			error(err.message);
-			debug(`Secret upload error stack: ${err.stack}`);
+			debug(`wrangler secret:bulk error stack: ${err.stack}`);
 		}
 		throw new Error(`Failed to upload secrets.`);
 	} finally {
--- a/src/index.ts
+++ b/src/index.ts
@@ -208,7 +208,7 @@ async function uploadSecrets() {
 	} catch (err: unknown) {
 		if (err instanceof Error) {
 			error(err.message);
-			debug(`wrangler secret:bulk error stack: ${err.stack}`);
+			err.stack && debug(err.stack);
 		}
 		throw new Error(`Failed to upload secrets.`);
 	} finally {
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,6 +5,7 @@ import {
 	endGroup as originalEndGroup,
 	error as originalError,
 	info as originalInfo,
+	debug,
 	startGroup as originalStartGroup,
 	setFailed,
 	setOutput,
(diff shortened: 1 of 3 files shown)


petebacondarwin:
Does this actually resolve the error surfacing? I have a feeling that Wrangler is swallowing the error before it even gets to the wrangler-action.

AdiRishi:
> Does this actually resolve the error surfacing? I have a feeling that Wrangler is swallowing the error before it even gets to the wrangler-action.

You're right in the sense that if wrangler cli itself is swallowing the error, this won't help.
However if that's not the case, at the very least this will log what we get back from the CLI output. Right now, even that error is swallowed and we just throw a generic `Failed to upload secrets.` error.

AdiRishi:
> I think in this case the error will be swallowed by `@actions/exec`, which will throw an error like `The process npx failed with exit code 1`. We need to add listeners to capture output from the running command as described here: https://github.com/actions/toolkit/blob/main/packages/exec/README.md#outputoptions

Oh I see! Thank you, was not aware of this. I'll update the PR

1000hz:
Sorry, I got mixed up. Adding listeners to capture output from `exec()` shouldn't be necessary to pipe the results to stdout/stderr of the parent process, since that's already the default behavior. I suspect the error output is indeed being swallowed by wrangler itself.

src/index.ts
```diff
@@ -204,7 +205,11 @@ async function uploadSecrets() {
 				),
 			),
 		});
-	} catch (err) {
+	} catch (err: unknown) {
+		if (err instanceof Error) {
+			error(err.message);
+			debug(`wrangler secret:bulk error stack: ${err.stack}`);
```
1000hz, line 211:
```suggestion
			debug(err.stack);
```

1000hz (commented):
I'm a little suspicious that `exec()` isn't actually executing during some of the reports of failed secret uploads, so I think this PR will be helpful to narrow down what's causing things to fail in case it's indeed happening before we actually execute wrangler.

Please add a changeset by running `npx changeset` describing this change.

AdiRishi:
Thanks again for the CR, definitely learned a few things about exec and it's output. If you'd like I can replace `exec` with `getExecOutput` and we can add stdout/stderr to the debug log as well.

1000hz (approved):
Thanks for the PR!

Issue #209, opened 2023-11-17
Wrangler 3.16 breaking secret uploads


demosjarco:
The action fails immediately since the release of Wrangler [3.16](https://github.com/cloudflare/workers-sdk/releases/tag/wrangler%403.16.0)
```
Finished processing secrets JSON file:
  ✨ 0 secrets successfully uploaded
  
  ✘ [ERROR] 🚨 23 secrets failed to upload
```
I assume it was [#4179](https://github.com/cloudflare/workers-sdk/pull/4179) that was merged that broke it

mrbbot:
Hey! 👋 I'm unable to reproduce this. Are you able to share the workflow run that is failing, or is it in a private repo? If so, could you share the full logs for the workflow?

demosjarco:
Looking back at the the log for the `3.16` fail, that was it. I don't see any other error message or log from wrangler. I see the secrets were assigned correctly. I have dm-ed you on discord the log file.

demosjarco:
Ok, updating to wrangler `3.17` that came out earlier today, it works (no change done to the secrets vs `3.15` that was working, `3.16` that wasn't, and `3.17` that is).

Side node, log format is broken, but I'm not sure if I should post that in this issue or start a new one.

leyafo:
I have the same problem that struggled me a day. I have set the CLOUDFLARE_API_TOKEN into my repository secrets settings. But can not get.
here is my configurations:
```
  - name: Upload to Cloudflare
        uses: cloudflare/wrangler-action@v3
        with:
          secrets: |
            CLOUDFLARE_API_TOKEN
            CLOUDFLARE_ACCOUNT_ID
          shell: 'bash'
          accountId: '${{ secrets.CLOUDFLARE_ACCOUNT_ID }}'
          apiToken: '${{ secrets.CLOUDFLARE_API_TOKEN }}'
          environment: ${{ github.events.inputs.environment }}
          command: 'whoami --env ${{ github.events.inputs.environment }}' 
          postCommands: |
            echo '${{ secrets.CLOUDFLARE_ACCOUNT_ID }}'
            echo ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
            echo "Uploading ${{ env.file }} to Cloudflare slippod/${{ env.filename }}..."
            wrangler r2 object put slippod/${{ env.filename }}) --file="${{ env.file }}"
            echo "***done***" 
        env:
[...]

PR #235, opened 2024-02-09
(feat): Use existing wrangler installation when appropriate

### LINKED ISSUE #199: Wrangler is reinstalled even when it's already available
Currently, `wrangler-action` always installs wrangler as specified by the `wranglerVersion` input or the `DEFAULT_WRANGLER_VERSION` constant contained in the source. This isn't always necessary as Wrangler may have already been installed in a previous step of the workflow (e.g. if it's declared in the project's dependencies in `package.json`). This also makes it harder to keep the version of Wrangler used by `wrangler-action` and the version used locally consistent, as the user needs to manually keep `wranglerVersion` and `package.json` in sync.

Instead, we should only install Wrangler if it's not already present in the `$PATH` (including the package manager bin directory) OR if `wranglerVersion` has been explicitly provided

AdiRishi:
Resolves #199 #259 

### Rationale

The code added to this PR has three notable behaviours
1. The `installWrangler` function now checks for existing wrangler installation. It does so by running the `wrangler --version` command
2. If the command is successful and the version returned is `exactly equal` to the required version as per config, the wrangler installation step is skipped
3. If no wrangler version was provided to the config of the action, and a pre-installed wrangler exists, the action will use the pre-installed wrangler
4. The above two steps are wrapped in a `try/catch` so as to prevent any negative behaviour from this change

The decision to check versions seems sound to me, as it is entirely conceivable that users of this library want to install a newer wrangler version in CI than the pinned version in their repositories.

Just to remind everyone, `wrangler --version` output can look like this
```
 ⛅️ wrangler 3.27.0 (update available 3.28.0)
-------------------------------------------------------
```

### Testing

- I've added a new folder in tests to run this action against a directory with pre-installed wrangler. This test won't fail if something goes wrong, but we can eye it's output to check things work.
- I know the above test method is a bit sub-par. Happy to hear more ideas on how we can test better. There are a few options here, but not sure how detailed we want to go.


Cherry:
> * If the command is successful and the version returned is `greater than or equal` to the required version as per config, the wrangler installation step is skipped

This specific part of this addition concerns me, as it would be a breaking change.

Today, specifying `wranglerVersion` in the config is a hard-set version, and not just a `semverGt` check. As a real-world example, I use the latest wrangler `3.28.x` when developing locally, but pin my version in CI to `3.15.0` to workaround secrets issues as per https://github.com/cloudflare/wrangler-action/issues/209. I believe this specific problem is resolved now, but it's not uncommon to want to pin to a specific and reproducible version in CI, even if that version when developing locally is a little more flexible.

AdiRishi:
Apologies for the delay in responding, got busy with something else.

@Cherry thanks for the comments, I didn't consider the exact version requirements for wrangler.
Due to the fact that this action pins a "default" wrangler version, I didn't want the behavior to check the pre-installed version against it. That would result in users of this action having to constantly update and keep the `wranglerVersion` config in sync with whatever they have in package.json

To solve this, I've added some detection in the code to see if a user provided `wranglerVersion` or if we're using the default. - If no wrangler version is provided and there exists a pre-installed wrangler in the folder, the action will now use the pre-installed version.
- If a wrangler version was specified, the action will check for an exact version match or else install the requested version

AdiRishi:
Hello 👋 
It's been a while since the last update, just want to bump this. Is anyone looking at this PR?

petebacondarwin:
Hey @AdiRishi - sorry about the radio silence. We have been rammed preparing for this weeks DevWeek announcements.
It will be back to business as usual next week, so I hope we can get someone to move this forward for you then.

AdiRishi:
@petebacondarwin a gentle bump again on this PR 🙏

petebacondarwin:
> Hmm, I looked into the test failure. It seems unrelated to my change? From looking through the code and the action output I would say that the `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ACCOUNT_ID` secrets have been accidentally removed from this repository.
> 
> We can see that in the `./run` output
> 
> ```
>  with:
>     workingDirectory: ./test/environment
>     environment: dev
>     preCommands: npx wrangler deploy --env dev
>     secrets: SECRET1
>   SECRET2
> ```
> 
> It doesn't list the `accountId` or `apiToken` field. What we would expect to see if they were present is something like this
> 
> ```
>  with:
>     accountId: ***
>     apiToken: ***
> ```

We can't run the full tests on a PR that comes from a 3rd party fork of the repository. Github Actions blocks sharing secrets with 3rd parties.

src/index.ts
```diff
@@ -82,6 +85,53 @@ async function installWrangler() {
...
+			throw new Error(
+				`Unable to parse Wrangler version from the output: ${stdout}`,
+			);
+		}
+		installedVersion = versionMatch[1];
+		if (config.didUserProvideWranglerVersion) {
+			if (semverEq(config.WRANGLER_VERSION, installedVersion)) {
+				info(`✅ Using Wrangler ${installedVersion}`, true);
+				endGroup();
+				return;
```
petebacondarwin, line 110: This nested return makes it a bit difficult to follow the flow here.
I think it would be cleaner if we use this block just to try identifying a currently installed version.
Then another block to determine whether we should use it or not.
And then finally a block that will install a version of we can't use the current version.

test/pre-installed-wrangler/package.json
```diff
@@ -0,0 +1,8 @@
+{
+	"name": "wrangler-action-pre-installed-wrangler-test",
+	"license": "MIT",
+	"private": true,
+	"devDependencies": {
+		"wrangler": "^3.28.0"
```
petebacondarwin, line 6: Since this is a test of checking to see if we can use a preinstalled version then it would make sense (with or without a package-lock file) to make this an explicit version.
```suggestion
		"wrangler": "3.28.0"
```

petebacondarwin: Actually how about we do something really crazy and build a fake wrangler into a `node_modules` directory here that responds to the `version` command with a hardcoded version and also supports some fake Wrangler command such as `test-wrangler-action`. If we end up installing a different version over this then running that command would fail.

petebacondarwin (changes requested):

test/pre-installed-wrangler/package.json
AdiRishi: That is truely a wild idea 🤯 
Any ideas on how to do something like this? The only thing that comes to my mind is something like [patch-package](https://www.npmjs.com/package/patch-package).
Would love to know your thoughts.

AdiRishi:
Thank you for the review! Some great points, I'll address them soon.

test/pre-installed-wrangler/package.json
petebacondarwin: Nothing so clever. I can push a commit for you to consider if you like

petebacondarwin: I pushed a commit but it didn't quite work in the CI as hoped. I can check again later today.

AdiRishi: Very interesting idea, I understand the path you were suggesting now. I've pushed a minor fix that I think will help its run in CI. Its working locally for me at least when I run `npx wrangler action-test`.

petebacondarwin:
OK things are looking good. Now that there is logic to reuse current installations of Wrangler, we need to be careful to remove any from tests to avoid leaking between them. E.g. https://github.com/cloudflare/wrangler-action/actions/runs/8970219683/job/24633372708#step:13:17

AdiRishi:
Understood, I'll take a pass through the existing tests and update them as needed.
Also sorry, I clearly forgot to run formatting in my last few commits haha 😅

AdiRishi:
@petebacondarwin I was looking through the tests, and we clearly reuse `./tests/base` and `tests/empty` a few times.
In my mind there are two main ways to tackle this.
1. Explicitly deal with the side effects, e.g `rm -rf node_modules` explicitly as part of our CI steps
2. Enforce a convention of having a different folder in `tests` for each step/check we want to do in CI. So we would have things like `tests/only-build`, `tests/build-quiet` etc etc

I think I prefer option two even though it's more duplication it seems like a better convention. What are your thoughts?

petebacondarwin:
> @petebacondarwin I was looking through the tests, and we clearly reuse `./tests/base` and `tests/empty` a few times. In my mind there are two main ways to tackle this.
> 
> 1. Explicitly deal with the side effects, e.g `rm -rf node_modules` explicitly as part of our CI steps.
> 2. Enforce a convention of having a different folder in `tests` for each step/check we want to do in CI. So we would have things like `tests/only-build`, `tests/build-quiet` etc etc
> 
> Imo option one adds more mental overhead when writing and running tests. I think I prefer option two even though it's more duplication it seems like a better convention. What are your thoughts?

I think I agree with you in this case. There is a trade-off if the cost of setting up an environment incerases which can cause CI jobs to slow down, but I think we are a long way off here.

AdiRishi:
Alright, gone through all the tests, setup new test folders where necessary. Hope this runs fine in CI 🤞

.github/workflows/deploy.yml
```diff
@@ -52,28 +51,20 @@ jobs:
...
           SECRET1: ${{ secrets.SECRET1 }}
           SECRET2: ${{ secrets.SECRET2 }}
 
-      - name: Clean up Deployed Environment Worker
-        uses: ./
-        with:
-          workingDirectory: "./test/base"
-          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
-          command: delete --name wrangler-action-dev-environment-test --force
```
petebacondarwin, line 68: Is this actually a test we want to do? i.e. run the delete command?

petebacondarwin (approved):
OK I think we are good here. I'll just try running it on a local branch and then merge.
Thanks for the iterations!

.github/workflows/deploy.yml
AdiRishi: That's a great question. We actually do have a `delete` command below when cleaing up the secret test workers. I figured this change along with another one down below would give us the chance to test out `postCommands`. Since there wasn't a previous use of it in this script.

.github/workflows/deploy.yml
```diff
@@ -52,28 +51,20 @@ jobs:
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
           environment: dev
           preCommands: npx wrangler deploy --env dev # https://github.com/cloudflare/wrangler-action/issues/162
+          postCommands: npx wrangler delete --name wrangler-action-dev-environment-test --force
```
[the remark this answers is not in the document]
AdiRishi: Yeah that sounds like something we should add. Do you want me to do it in this PR? If not I'm happy to make another PR after this one to add it in.

AdiRishi:
> OK I think we are good here. I'll just try running it on a local branch and then merge. Thanks for the iterations!

Awesome! Will you be pushing these commits to #260 ? I'm keen to see if it runs correctly myself :)

Issue #232, opened 2024-01-26
Update docs to avoid Node16 deprecation warning


Soviut:
The docs currently have an example that has

```yaml
uses: actions/checkout@v3
```

This gives a deprecation warning that Node16 should be replaced with Node20

<img width="590" alt="image" src="https://github.com/cloudflare/wrangler-action/assets/226882/20e86c25-fb21-4572-8394-a933a022a3fd">

To avoid steering the user into using a deprecated version we should change it to

```yaml
uses: actions/checkout@v4
```

PR #236, opened 2024-02-10
Update README.md to remove deprecation warnings

boidolr:
`actions/checkout@v3` runs on node16 which is sunset by GitHub Action runners.

resolves https://github.com/cloudflare/wrangler-action/issues/232

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Deploy
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Deploy
         uses: cloudflare/wrangler-action@v3
         with:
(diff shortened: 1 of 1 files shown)


petebacondarwin (changes requested):
This causes the CI tests to fail. We'll need these to pass before we can land this.

boidolr:
> This causes the CI tests to fail. We'll need these to pass before we can land this.

I am not clear on what to change here. The README.md changed, and now a build complains about `CLOUDFLARE_API_TOKEN` not being set when building. Any guidance on what exactly needs to change?

petebacondarwin:
Doh! Sorry. My mistake. Since the PR is coming from your fork, you don't have access to the secrets in Github for the action to run successfully.

PR #242, opened 2024-03-10
docs: add warning about packageManager config

capaj:
just something I bumped into today

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ jobs:
       - name: Deploy
         uses: cloudflare/wrangler-action@v3
         with:
+          packageManager: pnpm # you can omit this if you use npm
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
 ```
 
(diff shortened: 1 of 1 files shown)


petebacondarwin (approved):
I suspect we should be able to infer the package manager in use??

Cherry:
@petebacondarwin This is inferred by https://github.com/cloudflare/wrangler-action/pull/193, and I'm not really sure the "default" copy/pasted config should use `pnpm` - that's definitely not the standard.

> "The package manager you'd like to use to install and run wrangler. If not specified, the preferred package manager will be inferred based on the presence of a lockfile or fallback to using npm if no lockfile is found. Valid values are `npm` | `pnpm` | `yarn` | `bun`."

You do not need to specify `packageManager` as the docs now suggests, as long as you're running an up to date version.

Cherry:
The detection of `pnpm` looks for a `pnpm-lock.yaml` file in the working directory, but only after looking for a `package-lock.json`, and then a `yarn.lock`.

If you have a `pnpm-lock.yaml` committed, and none of these other files, it should work as expected.

README.md
```diff
@@ -52,6 +53,10 @@ jobs:
 
 ## Configuration
 
+### packageManager
```
IgorMinar, line 56: Unfortunately this header makes all the following paragraphs appear as if they were part of the packageManager docs.

PR #255, opened 2024-04-26
Stop racing secret uploads

matthewdavidrodgers:
For up to date versions of wrangler, secrets are uploaded via the 'secret:bulk' command, which batches updates in a single API call.

For versions of wrangler without that capability, the action falls back to the single 'secret put' command for each secret. It races all these with a Promise.all()

Unfortunately, the single secret API cannot handle concurrency - at best, these calls have to wait on one another, holding requests open all the while. Often it times out and errors.

This fixes the legacy secret upload errors by making these calls serially instead of concurrently.

### DIFF
--- a/.changeset/funny-boats-care.md
+++ b/.changeset/funny-boats-care.md
@@ -1,0 +1,13 @@
+---
+"wrangler-action": minor
+---
+
+Stop racing secret uploads
+
+For up to date versions of wrangler, secrets are uploaded via the 'secret:bulk' command, which batches updates in a single API call.
+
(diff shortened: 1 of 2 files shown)


taylorlee (approved):

rozenmd (approved):

RamIdeas (approved):

RamIdeas:
I think this change makes sense. If you're on a legacy version of wrangler, pre-`secret:bulk`, you'll get a slower but more stable experience. You'll need to upgrade to newer versions of wrangler, post-`secret:bulk`, to get the faster bulk upload.

Tests were run in #256, since this PR came from a fork.

Issue #194, opened 2023-10-10
Refactor tests to be integration tests rather than E2E deployments


1000hz:
The way this action is currently being tested is primarily through a GitHub Actions [workflow](https://github.com/cloudflare/wrangler-action/blob/0e6f0d30805b05d6dbd8861f317900d045526d80/.github/workflows/deploy.yml) that deploys several Worker fixtures and passes as long as the action doesn't fail to complete.

This approach has a few severe drawbacks. Because deploying these Workers to our test account requires our API token, PRs submitted by external contributors don't have access to our secrets and these always fail to run. These failures also mask those that might potentially occur during other checks like code formatting. This results in us merging sometimes-broken code due to having no reliable confidence signal before merging.

Second, the current E2E structure makes it difficult to assert against expected behaviors of the action in an automated way. It's also cumbersome to construct tests for each scenario (and tear them down after the test completes, etc.) These E2E tests end up mostly testing the underlying behavior of wrangler rather than the behavior of the action itself.

We can improve our test coverage by moving towards integration testing for much of the behavior currently being tested (to varying degrees) in an E2E manner. To do this, we'll invoke the action's entrypoint in a vitest and assert that the expected subprocesses are spawned appropriately.

We can limit E2E tests to a few basic smoke tests that only run `if: ${{ github.repository_owner == 'cloudflare' }}` while allowing the rest of the test suite to run on external PRs.

PR #271, opened 2024-06-16
Unreverts #235 and don't automatically install wrangler when checking if it present

Maximo-Guk:
See [!239](https://github.com/cloudflare/wrangler-action/pull/269) for more context

This PR adds the `execNoInstall` field to packageManagers, so that when we're executing wrangler to check if it present with our various package manager configurations, we don't automatically install it, in the case that it isn't present.

This was tested by running [test-fixup-235 branch](https://github.com/Maximo-Guk/wrangler-action/tree/maximo/test-fixup-235) on the [test-action](https://github.com/Maximo-Guk/demo-actions/tree/maximo/test-action) branch of my demo-actions repo which contains hello-world workers applications for npm, pnpm, yarn and bun.

The successful pipeline where each package manager was tested https://github.com/Maximo-Guk/demo-actions/actions/runs/9533218220

cc @AdiRishi


AdiRishi:
@Maximo-Guk as I said in the comment in my PR, this approach is much better. I like how you solved it 🚀

AdiRishi (approved):

AdiRishi (commented):
The only thing I'll add is that it might be good to add a test to ensure it installs the correct version of wrangler by default. Could be a test we add directly into `deploy.yml`, perhaps a simple exection of `wrangler --version` and checking that it returns the exected output.

jahands (approved):

Maximo-Guk:
> The only thing I'll add is that it might be good to add a test to ensure it installs the correct version of wrangler by default. Could be a test we add directly into `deploy.yml`, perhaps a simple exection of `wrangler --version` and checking that it returns the exected output.

I'd really like to add a test for this, however with the current state of tests in this repo #194 I'm going to hold off on that for now, and get this merged in relying on the manual tests.

I'm hoping to work on converting most of the e2e `deploy.yaml` tests to proper integration tests sometime soon

Issue #277, opened 2024-07-11
Install version of wrangler not being used with v3.7.0


acusti:
i’m very excited for the changes introduced in v3.7.0 to use existing wrangler installation when appropriate (from #271). my app relies on `getPlatformProxy` so it cannot be deployed with any version of wrangler that doesn’t include that API. however, in my case, the logic to check that existing install and version if somehow falling through to the `⚠️ Wrangler not found or version is incompatible. Installing...` [fallback logic](https://github.com/cloudflare/wrangler-action/pull/271/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R133-R136).

here is my output from the workflow run, which shows that i am using yarn v1.22.22 and shows that `/usr/local/bin/yarn wrangler --version`, as invoked by wrangler-action, is returning the version number (`3.63.2`) correctly:

```
Run cloudflare/wrangler-action@v3.7.0
  with:
    apiToken: ***
    quiet: false
🔍 Checking for existing Wrangler installation
  /usr/local/bin/yarn wrangler --version
  yarn run v1.22.22
  $ /home/runner/work/outlyne/outlyne/node_modules/.bin/wrangler --version
  3.63.2
  Done in 0.57s.
  ⚠️ Wrangler not found or version is incompatible. Installing...
📥 Installing Wrangler
  /usr/local/bin/yarn add wrangler@3.13.2
  yarn add v1.22.22
  [1/5] Validating package.json...
  [2/5] Resolving packages...
  [3/5] Fetching packages...
  [4/5] Linking dependencies...
```

my hypothesis is that the issue is due to the fallback regex starting with the “start of string” token (`/^`):
```js
stdout.match(/wrangler (\d+\.\d+\.\d+)/) ?? stdout.match(/^(\d+\.\d+\.\d+)/);
```

i’m not sure how `stdout` is derived, but if it includes anything before the version output (e.g. `\n3.63.2`, or even `/home/runner/work/outlyne/outlyne/node_modules/.bin/wrangler --version\n3.63.2`), the regex won’t match it.

the most obvious solution, then, is to use multi-line mode so that `^` matches the start of every line as well as the start of the string:

```js
		const versionMatch =
 			stdout.match(/wrangler (\d+\.\d+\.\d+)/) ??
 			stdout.match(/^(\d+\.\d+\.\d+)/);
```

--------

in case it’s helpful, this is my yaml from the workflow file itself:

```yml
    deploy:
[...]

PR #278, opened 2024-07-11
When checking for existing wrangler install, match wrangler version even if multiline output

acusti:
implements #277

i don’t know how to test or verify this, but in case it’s helpful, here’s my proposed solution (assuming my hypothesis from #277 is correct).

i think this will also fix the issue documented here: https://github.com/cloudflare/wrangler-action/issues/199#issuecomment-2188863592

### DIFF
--- a/.changeset/gentle-cobras-move.md
+++ b/.changeset/gentle-cobras-move.md
@@ -1,0 +1,5 @@
+---
+"wrangler-action": patch
+---
+
+fix: Detect existing wrangler install even when wrangler version output is multiline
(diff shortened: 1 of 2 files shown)


austinorth (approved):
Looks good to me!

austinorth:
@Maximo-Guk Can we get this merged please?

threepointone (approved):

Maximo-Guk:
Hi @acusti this looks good to me, thank you! Could you add a changeset please? 

For [reference](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets)

acusti:
@Maximo-Guk done!

Maximo-Guk (approved):
Thank you!

PR #294, opened 2024-09-26
chore: add support for new bun.lock

Cherry:
Adds support for upcoming `bun.lock` to  `detectPmMismatch ` method. 

https://x.com/jarredsumner/status/1839056611765268651


Maximo-Guk (approved):
Thank you!

PR #295, opened 2024-09-27
WO-202: update wrangler version so the github action works out of the box with Workers Observability

Ankcorn:
Hi folks, several users using this action have reported observability enabled=true not working in there wrangler configs

Tracked the issue down to the default version here being out of date.

I would be grateful if we could bump this up to the latest


ashkulz:
Or at the very least, allow using `latest` in the action configuration -- it's difficult as Dependabot won't bump this up automatically in our actions.

penalosa (approved):

ashkulz:
@penalosa can this make way into a `v3.8.1` release?

PR #303, opened 2024-10-18
Add parity with pages-action for Pages deploy outputs

courtney-sims:
Enables Pages deploys to provide outputs for id, environment, url, and alias, just like they can in pages-action. The current way of getting deploy outputs with Workers is via wrangler output to std out, but this information for Pages was not available in std out.

Because of that, this change adds a new method for getting outputs via wrangler artifacts.

The new outputs look like this in the Actions view.

When an alias is not available in the artifact:
Screenshot 2024-10-18 at 2 38 10 PM

When an alias is available in the artifact:
Screenshot 2024-10-18 at 2 41 05 PM

When we don't get stuff back from the request for artifact:
Screenshot 2024-10-21 at 10 14 05 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
 	},
 	"dependencies": {
 		"@actions/core": "^1.10.1",
-		"@actions/exec": "^1.1.1"
+		"@actions/exec": "^1.1.1",
+		"zod": "^3.23.8"
 	},
 	"devDependencies": {
 		"@changesets/changelog-github": "^0.4.8",
@@ -39,6 +40,7 @@
 		"@types/node": "^20.10.4",
 		"@vercel/ncc": "^0.38.1",
 		"prettier": "^3.1.0",
+		"mock-fs": "^5.4.0",
 		"semver": "^7.5.4",
 		"typescript": "^5.3.3",
 		"vitest": "^1.0.3"
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,11 +1,11 @@
 import {
+	debug,
 	getBooleanInput,
 	getInput,
 	getMultilineInput,
 	endGroup as originalEndGroup,
 	error as originalError,
 	info as originalInfo,
-	debug,
 	startGroup as originalStartGroup,
 	setFailed,
 	setOutput,
@@ -13,10 +13,11 @@ import {
 import { getExecOutput } from "@actions/exec";
 import semverEq from "semver/functions/eq";
 import { exec, execShell } from "./exec";
-import { checkWorkingDirectory, semverCompare } from "./utils";
 import { getPackageManager } from "./packageManagers";
+import { checkWorkingDirectory, semverCompare } from "./utils";
+import { getDetailedPagesDeployOutput } from "./wranglerArtifactManager";
 
-const DEFAULT_WRANGLER_VERSION = "3.78.10";
+const DEFAULT_WRANGLER_VERSION = "3.81.0";
 
 /**
  * A configuration object that contains all the inputs & immutable state for the action.
@@ -313,6 +314,9 @@ async function wranglerCommands() {
 			let stdErr = "";
 
 			// Construct the options for the exec command
+			const wranglerOutputDir = "/opt/wranglerArtifacts";
+			process.env.WRANGLER_OUTPUT_FILE_DIRECTORY = wranglerOutputDir;
+
 			const options = {
 				cwd: config["workingDirectory"],
 				silent: config["QUIET_MODE"],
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -39,6 +39,7 @@
 		"@types/node": "^20.10.4",
 		"@vercel/ncc": "^0.38.1",
 		"prettier": "^3.1.0",
+		"mock-fs": "^5.4.0",
```
courtney-sims, line 42: Used by tests to mock the filesystem, but open to another approach for this that doesn't involve adding a new package!

courtney-sims: I thought this approach was a really simple and readable one though

src/index.ts
```diff
@@ -357,6 +356,22 @@ async function wranglerCommands() {
...
 			}
+			// Check if this command is a pages deployment
+			if (
+				command.startsWith("pages publish") ||
+				command.startsWith("pages deploy")
+			) {
+				const pagesArtifactFields =
+					await getDetailedPagesDeployOutput(wranglerOutputDir);
+
+				if (pagesArtifactFields) {
```
Maximo-Guk, line 367: Maybe if `pagesArtifactFields` is null, we can log a message that outputs haven't been set, maybe due to being on an old wrangler version?

package.json
Maximo-Guk: This is awesome! Since it's just used in tests, we can probably make it a `devDependencies`?

src/wranglerArtifactManager.ts
```diff
@@ -0,0 +1,80 @@
+import { open, readdir } from "fs/promises";
+
+interface OutputEntryBase<T extends string> {
+	version: number;
+	type: T;
+}
+
+export interface OutputEntryPagesDeployment
```
Maximo-Guk, line 8: Can we use zod here?

courtney-sims: Ah! This was a comment a I mean to leave. We aren't using zod anywhere else in this repo, so I wasn't sure if that would be appropriate. You think we should start using it?

package.json
courtney-sims: I think it is, right?

src/wranglerArtifactManager.ts
Maximo-Guk: I think so! It would be really great for catching if the schema changes at all for some bizarre reason

package.json
Maximo-Guk: Oops! I was not paying attention!!

src/index.ts
Maximo-Guk: Hmm, I was thinking an info log like we do in a couple places already https://github.com/cloudflare/wrangler-action/blob/main/src/index.ts#L118

What do you mean by logging in the deployment details?

src/wranglerArtifactManager.ts
courtney-sims: Cool. Will do

src/wranglerArtifactManager.ts
```diff
@@ -0,0 +1,80 @@
...
+		for await (const line of file.readLines()) {
+			try {
+				const output = JSON.parse(line);
+				const parsedOutput = OutputEntryPagesDeployment.parse(output);
+				if (parsedOutput.type === "pages-deploy-detailed") {
+					// Assume, in the context of the action, the first detailed deploy instance seen will suffice
+					return parsedOutput;
+				}
+			} catch (err) {
+				// If the line can't be parsed, skip it
```
Maximo-Guk, line 71: Should we also log a warning if we can't parse a line here?

courtney-sims: Is that something the user should see? That would be a problem between our code and wrangler's

courtney-sims: Do we have internal logs?

Maximo-Guk: I think it would be cool if we could record this information in to sentry somehow, like workers-sdk does, but that would be a pretty big lift since we don't already have sentry configured in this project. For now exposing this to the user with some sort of message would probably be good enough I think, so we don't silently skip and swallow the exception here

courtney-sims: We discussed this more in chat and decided to just stick with the message about the version update in index due to the complexity of the situation.

Maximo-Guk (approved):

src/wranglerArtifactManager.test.ts
```diff
@@ -0,0 +1,87 @@
+import { describe, expect, it } from "vitest";
+import {
+	getDetailedPagesDeployOutput,
+	getWranglerArtifacts,
+} from "./wranglerArtifactManager";
+
+describe("wranglerArtifactsManager", () => {
+	const mock = require("mock-fs");
```
Maximo-Guk, line 8: We should probably be using an es6 import here?

src/wranglerArtifactManager.test.ts
```diff
@@ -0,0 +1,87 @@
...
+					"not-wrangler-output.json": "test",
+				},
+			});
+
+			const artifacts = await getWranglerArtifacts("./testOutputDir");
+
+			expect(artifacts).toEqual([
+				"./testOutputDir/wrangler-output-2024-10-17_18-48-40_463-2e6e83.json",
+			]);
+			mock.restore();
```
Maximo-Guk, line 26: We could probably move `mock.restore()` in to an `afterEach()`

courtney-sims: Good point

Issue #306, opened 2024-10-24
Release 3.10.0 fails depending on the file system permissions


adriantrunzo:
# Description

I've started seeing the following failures with the latest release:

```
✘ [ERROR] EACCES: permission denied, mkdir '/opt/wranglerArtifacts'
```

The failures started at the same time as the 3.10.0 release. Indeed, that release references the `/opt/wranglerArtifacts` directory: https://github.com/cloudflare/wrangler-action/commit/3ec7f8943ef83351f743cfaa8763a9056ef70993#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R317.

I do not have the ability to change permissions in these runners. Setting `WRANGLER_OUTPUT_FILE_DIRECTORY` didn't have any effect.

# Expected Behavior

No failure.

# Possible Solution

- Simply ignore the new functionality if a file system error occurs
- Add a configuration option for the output directory
- Use `WRANGLER_OUTPUT_FILE_DIRECTORY`

JadCham:
I was about to submit an issue for the exact same error. 



**Workaround**
Pin version to cloudflare/wrangler-action@v3.9.0

**What we tried**
Pinning `wranglerVersion` only to `3.78.10` in order to get around the `DEFAULT_WRANGLER_VERSION` bump but that did not work.
https://github.com/cloudflare/wrangler-action/pull/303/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80L19

Which points to the issue being in the action only as @adriantrunzo mentioned https://github.com/cloudflare/wrangler-action/commit/3ec7f8943ef83351f743cfaa8763a9056ef70993#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R317

adriantrunzo:
Thanks for sharing what you tried @JadCham. I tried with the latest version of `wrangler`, 3.12.0, in addition to an older one like you did.

I can also confirm that pinning the version to 3.9.0 is the workaround.

kashalls:
Running into this issue with self-hosted runner's as well.

```
🚀 Running Wrangler Commands
  /home/runner/_work/_tool/node/18.20.4/x64/bin/npx wrangler pages deploy --project-name=kubernetes-schemas --branch main .
  ✘ [ERROR] EACCES: permission denied, mkdir '/opt/wranglerArtifacts'
  ```

[...]

PR #312, opened 2024-10-31
Unrevert add parity with wrangler action

Maximo-Guk:
Unreverts #303 

Tested pipelines here:

- [Pipeline](https://github.com/Maximo-Guk/demo-actions/actions/runs/11607252464/job/32320464024#step:5:9) shows all outputs if wrangler version >=3.81.0
- [Pipeline](https://github.com/Maximo-Guk/demo-actions/actions/runs/11607284817/job/32320541183#step:5:9) falls back to only outputting `deployment-url` and `deployment-alias-url` if wrangler version doesn't output metadata file

Moves wrangler output to tmpdir rather than /opt/ since /opt/ is owned by root.

- Github self hosted runners may not have permissions to write to /opt/ #306 

- Also fallsback to trying to extract the deployment-url and deployment-alias-url from stdout when WRANGLER_OUTPUT_DIR is not specified

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,6 +16,8 @@ import { exec, execShell } from "./exec";
 import { getPackageManager } from "./packageManagers";
 import { checkWorkingDirectory, semverCompare } from "./utils";
 import { getDetailedPagesDeployOutput } from "./wranglerArtifactManager";
+import { join } from "path";
+import { tmpdir } from "os";
 
 const DEFAULT_WRANGLER_VERSION = "3.81.0";
 
@@ -34,6 +36,7 @@ const config = {
 	COMMANDS: getMultilineInput("command"),
 	QUIET_MODE: getBooleanInput("quiet"),
 	PACKAGE_MANAGER: getInput("packageManager"),
+	WRANGLER_OUTPUT_DIR: `${join(tmpdir(), "wranglerArtifacts")}`,
 } as const;
 
 const packageManager = getPackageManager(config.PACKAGE_MANAGER, {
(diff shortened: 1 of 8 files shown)


src/index.ts
```diff
@@ -333,27 +361,36 @@ async function wranglerCommands() {
...
+					config.WRANGLER_OUTPUT_DIR,
 				);
-				if (aliasUrlMatch && aliasUrlMatch.length == 2 && aliasUrlMatch[1]) {
-					const aliasUrl = aliasUrlMatch[1].trim();
+
+				if (pagesArtifactFields) {
+					setOutput("id", pagesArtifactFields.deployment_id);
+					setOutput("deployment-url", pagesArtifactFields.url);
+					setOutput("deployment-alias-url", pagesArtifactFields.alias);
+					setOutput("environment", pagesArtifactFields.environment);
```
Maximo-Guk, line 384: We need to support `deployment-url` and `deployment-alias-url` as this was what is [currently used](https://github.com/cloudflare/wrangler-action?tab=readme-ov-file#using-the-deployment-url-and-deployment-alias-url-output-variables) in wrangler-action.

Need to figure out if `id` and `environment` are what we want to go with ( this is what [pages-action](https://github.com/cloudflare/wrangler-action?tab=readme-ov-file#using-the-deployment-url-and-deployment-alias-url-output-variables )) uses. However, we've been prefixing outputs with `deployment` in wrangler-action

src/index.ts
```diff
@@ -33,6 +36,7 @@ const config = {
 	COMMANDS: getMultilineInput("command"),
 	QUIET_MODE: getBooleanInput("quiet"),
 	PACKAGE_MANAGER: getInput("packageManager"),
+	WRANGLER_OUTPUT_DIR: `${join(tmpdir(), "wranglerArtifacts")}`,
```
Maximo-Guk, line 39: Use tmpdir for wranglerArtifacts

src/index.ts
```diff
@@ -276,6 +280,28 @@ async function uploadSecrets() {
 	}
 }
 
+function extractDeploymentUrlsFromStdout(stdOut: string): {
```
jrf0110, line 283: Wait why are parsing these things from stdout? This should be in wrangler artifacts, no?

Maximo-Guk: Added a comment // fallback to trying to extract the deployment-url and deployment-alias-url from stdout for wranglerVersion < 3.81.0

PR #322, opened 2024-11-07
Add tests for installWrangler

courtney-sims:
Baby step refactor to get more tests into this repo:
- isolates index functions from side effects
- creates type for config
- uses config and package manager types for parameters to index functions so we can pass in some test data
- adds some tests for the InstallWrangler method

Tested by running this version on a demo site:
Screenshot 2024-11-07 at 3 47 49 PM


src/wranglerAction.ts
```diff
@@ -0,0 +1,455 @@
...
+import { getExecOutput } from "@actions/exec";
+import semverEq from "semver/functions/eq";
+import { z } from "zod";
+import { exec, execShell } from "./exec";
+import { PackageManager } from "./packageManagers";
+import { semverCompare } from "./utils";
+import { getDetailedPagesDeployOutput } from "./wranglerArtifactManager";
+
+export type wranglerActionConfig = z.infer<typeof wranglerActionConfig>;
+export const wranglerActionConfig = z.object({
```
jrf0110, line 20: Types are typically PascalCase

```suggestion
export type WranglerActionConfig = z.infer<typeof WranglerActionConfig>;
export const WranglerActionConfig = z.object({
```

jrf0110, line 20: Is this runtime type ever actually used? I don't see any `wranglerActionConfig.parse(...)` calls

src/wranglerAction.ts
```diff
@@ -0,0 +1,456 @@
...
+} from "@actions/core";
+import { getExecOutput } from "@actions/exec";
+import semverEq from "semver/functions/eq";
+import { z } from "zod";
+import { exec, execShell } from "./exec";
+import { PackageManager } from "./packageManagers";
+import { semverCompare } from "./utils";
+import { getDetailedPagesDeployOutput } from "./wranglerArtifactManager";
+
+export type WranglerActionConfig = z.infer<typeof wranglerActionConfig>;
```
jrf0110, line 19: Now that we're defining a type for config, we should use it when instantiating `config` in `index.ts` here https://github.com/cloudflare/wrangler-action/pull/322/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80L40

```tsx
/**
 * A configuration object that contains all the inputs & immutable state for the action.
 */
const config: WranglerActionConfig = {
	WRANGLER_VERSION: getInput("wranglerVersion") || DEFAULT_WRANGLER_VERSION,
	didUserProvideWranglerVersion: Boolean(getInput("wranglerVersion")),
	secrets: getMultilineInput("secrets"),
	workingDirectory: checkWorkingDirectory(getInput("workingDirectory")),
	CLOUDFLARE_API_TOKEN: getInput("apiToken"),
	CLOUDFLARE_ACCOUNT_ID: getInput("accountId"),
	ENVIRONMENT: getInput("environment"),
	VARS: getMultilineInput("vars"),
	COMMANDS: getMultilineInput("command"),
	QUIET_MODE: getBooleanInput("quiet"),
	PACKAGE_MANAGER: getInput("packageManager"),
	WRANGLER_OUTPUT_DIR: `${join(tmpdir(), "wranglerArtifacts")}`,
} ;
```

jrf0110 (approved):

src/wranglerAction.ts
courtney-sims: Good point! Thanks!

Maximo-Guk (approved):

.github/workflows/deploy.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: "latest"
+          node-version: "22.11.0"
```
Maximo-Guk, line 17: Maybe we could just pin it to 22?

Issue #301, opened 2024-10-15
[pages-action migration] Add Support for Deployments Feature 


aninibread:
**Context:** With the deprecation of [`pages-action`](https://github.com/cloudflare/pages-action), we are encouraging users who relied on it for Pages project deployments to migrate to `wrangler-action`.

**Problem:** We recognize that `wrangler-action` currently does not include the "Deployments" feature available in `pages-action`. We are actively working to add this feature to `wrangler-action` to align with `pages-action` capabilities and ensure users have the visibility and control they need when managing their deployments.

Please follow this issue for updates as we progress. We also welcome your feedback. If there are other deployment-related features or improvements you'd like to see, please share your suggestions here.

**From:** Cloudflare Workers Team

sebdanielsson:
Waiting for the next ansible-core release. Probably 2.18.1.

dargmuesli:
Maybe the README in https://github.com/cloudflare/pages-action can be updated to not point to this issue anymore (would require a short un-archival to push the change).

PR #325, opened 2024-11-13
Add GitHub deployments for parity with pages-action

Maximo-Guk:
Addresses #301 

This PR updates `wrangler-action` dependencies and adds `@actions/github`, so that we can create GitHub deployments, as well as job summaries when a user provides their GiHub token to the action config.

Github deployments successfully tested [here](https://github.com/Maximo-Guk/demo-actions/deployments/preview):
image

Job summary successfully tested [here](https://github.com/Maximo-Guk/demo-actions/actions/runs/11826667451/attempts/1#summary-32953204572)
image


src/wranglerAction.ts
```diff
@@ -424,6 +424,36 @@ async function wranglerCommands(
 					setOutput("pages-deployment-alias-url", pagesArtifactFields.alias);
 					setOutput("pages-deployment-id", pagesArtifactFields.deployment_id);
 					setOutput("pages-environment", pagesArtifactFields.environment);
+					// create github deployment, if GITHUB_TOKEN is provided
```
courtney-sims, line 427: Are github deploys only intended for pages? Is that not a thing for workers?

Maximo-Guk:
Do not merge until:

- [ ] wrangler has been released with new fields

src/wranglerAction.ts
```diff
@@ -424,6 +424,36 @@ async function wranglerCommands(
...
+					if (
```
courtney-sims, line 428: This logic for outputs is already not very friendly. I don't feel great about tangling up the github deployments with it. Could we add in some separation?

src/github.ts
```diff
@@ -0,0 +1,78 @@
+import { summary } from "@actions/core";
+import { context, getOctokit } from "@actions/github";
+import { env } from "process";
+import { WranglerActionConfig } from "./wranglerAction";
+
+type Octokit = ReturnType<typeof getOctokit>;
+
+export async function createGitHubDeployment({
```
courtney-sims, line 8: Is there a test we could write to make sure we're able to successfully call these octokit methods when we should?

src/github.ts
```diff
@@ -0,0 +1,78 @@
...
+			// don't have project_name or deployment_id I think
+			log_url: `https://dash.cloudflare.com/${config.CLOUDFLARE_ACCOUNT_ID}/pages/view/${projectName}/${deploymentId}`,
+			description: "Cloudflare Pages",
+			state: "success",
+			auto_inactive: false,
+		});
+	}
+}
+
+export async function createJobSummary({
```
courtney-sims, line 56: Similarly with this one, is there a test we could write to ensure the summary looks like we expect it to when we add our stuff to it?

src/wranglerAction.ts
Maximo-Guk: For now, we'll just have it set up for pages for pages-action parity, but we can definitely support workers in the future

src/wranglerAction.ts
```diff
@@ -15,7 +15,7 @@ import { exec, execShell } from "./exec";
 import { PackageManager } from "./packageManagers";
 import { semverCompare } from "./utils";
 import { getDetailedPagesDeployOutput } from "./wranglerArtifactManager";
-import { createGitHubDeployment, createJobSummary } from "./github";
+import { createGitHubDeployment, createJobSummary } from "./service/github";
```
courtney-sims: I don't see the service directory. Is this a magic thing?

courtney-sims: Oh. It's in a different commit. Nvm

src/service/github.ts
```diff
@@ -0,0 +1,117 @@
...
+		repo: context.repo.repo,
+		ref: githubBranch || context.ref,
+		auto_merge: false,
+		description: "Cloudflare Pages",
+		required_contexts: [],
+		environment,
+		production_environment: productionEnvironment,
+	});
+
+	if (deployment.status === 201) {
```
courtney-sims, line 40: Is there anything we wanna do if the status isn't a 201?

src/service/github.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+afterEach(() => {
+	mockfs.restore();
+});
+
+describe("github", () => {
+	it("Calls createGitHubDeployment successfully", async () => {
+		const githubUser = "mock-user";
+		const githubRepoName = "wrangler-action";
+		const server = setupServer(
+			...mockGithubDeployments({ githubUser, githubRepoName }).handlers,
```
courtney-sims, line 19: Is there anything we should assert on the mock?

src/test/test-utils.ts
```diff
@@ -0,0 +1,26 @@
+import { WranglerActionConfig } from "../wranglerAction";
+
+export function getTestConfig({
```
courtney-sims, line 3: testConfig didn't immediately hit in my mind as a test version of WranglerActionConfig when I saw it used in the test file. How do you feel about naming it more specifically testWranglerActionConfig? I know it's longer and that's basically the name of the repo, so I'm open to leaving it as-is if you feel strongly about it.

courtney-sims:
I like having all the github logic into one file!

src/test/test-utils.ts
Maximo-Guk: I honestly kinda like the shorter name! I think there's no other configs in the repo, it's probably fine?

src/service/github.spec.ts
Maximo-Guk: We are asserting on the request payload [here](https://github.com/cloudflare/wrangler-action/pull/325/files#diff-fd928f78f59ad2b55c465896ff5ebab1a82008825bae3460fb791dcf40efd873R16-R27), I suppose we could also check the response payload, but the response schema is [pretty long](https://docs.github.com/en/rest/deployments/deployments?apiVersion=2022-11-28#create-a-deployment), and we're not currently returning any values from [createGithubDeployment](https://github.com/cloudflare/wrangler-action/pull/325/files#diff-79d58800fd43beb67fbbc9bb955038da16724c49838b6f0d17a52a45458cda97R45).

I don't know if there's much value in mocking the response payload for those reasons

aaronadamsCA:
Great to see this happening!

I contributed some of the Pages action's features, and I have a few suggestions to avoid repeating some of its mistakes:

- **Don't require GitHub token as an action input.** Default its value to `github.token` ([see `actions/checkout` for a great example](https://github.com/actions/checkout/blob/cbb722410c2e876e24abbe8de2cc27693e501dcb/action.yml#L12-L24)), then let users turn features on/off using boolean inputs. This is a lot less confusing and aligns with the rest of the ecosystem.

- **Accept environment name as input.** Named GitHub environments can hold configurations, so migrating users might need this action's environment names to match their old ones, so that those configurations will still apply. And one repo can deploy multiple Wrangler apps, so their environment names must be configurable or else they will collide.

- **Provide separate inputs to enable deployments vs. job summaries.** Deployments add a lot of noise on the Conversation tab, whereas job summaries add a lot of noise on the Checks tab. Ideally users could choose how much noise they want.

src/test/test-utils.ts
jahands: yeah I think since we only have one get config function it's ok

jahands (approved):
LGTM!

src/service/github.spec.ts
courtney-sims: No, I think this is okay. All we want to do is test that we're reaching out the way we should be. We aren't doing anything with the response, so I don't think we need to test that it comes back a certain way. The request assertion seems good!

courtney-sims (approved):

Issue #240, opened 2024-03-05
Deployment failing due to secrets not uploading


vladinator1000:
I just migrated to wrangler-action v3 and this started happening to me. It fails every time I run it.

![image](https://github.com/cloudflare/wrangler-action/assets/1226564/0cac1064-ebfd-413c-801d-30572b78cb63)

Does anyone see where I'm going wrong? Here's my workflow file:
```yaml
name: Deploy

on:
  workflow_dispatch:
  push:
    branches: [main]

jobs:
  build_and_deploy:
    name: Build and deploy
    runs-on: ubuntu-latest

    strategy:
      matrix:
        node-version: [21.x]

    env:
      ENVIRONMENT: production
      APP_SECRET: ${{ secrets.APP_SECRET }}
      DATABASE_URL: ${{ secrets.DATABASE_URL }}
      CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}

    steps:
      - uses: actions/checkout@v4
      - name: Use Node.js ${{ matrix.node-version }}
        uses: actions/setup-node@v4
        with:
          node-version: ${{ matrix.node-version }}

      - uses: c-hive/gha-yarn-cache@v2

      - name: Get version
        id: version
        run: echo "::set-output name=version::$(date +'%Y-%m-%dT%H:%M:%S')-${{ github.sha }}"

      - name: Install dependencies
        run: yarn --frozen-lockfile

      - name: 🔨📦 Build and deploy
        uses: cloudflare/wrangler-action@v3
        with:
          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
          environment: 'production'
          secrets: |
            APP_SECRET
            DATABASE_URL
```

AdiRishi:
Hey 👋 
These issues are always quite tricky to debug so let me ask some simple debugging questions first
- Is your worker already deployed? Or is it yet to be created? If not, you should deploy it without secrets first
- Second, let's make sure the secret values are being passed in to your action, can you try to expand the group for `cloudflare/wrangler-action@3`, here's an example of what of of my projects looks like (note the inclusion of secret values in env)
    ```
    Run cloudflare/wrangler-action@v3
      with:
        accountId: ***
        apiToken: ***
        wranglerVersion: 3.26.0
        secrets: TURBO_TOKEN
        command: deploy --minify
        quiet: false
      env:
[...]

PR #328, opened 2024-11-18
fix: Use wrangler secret put instead of deprecated secret:bulk command

jahands:
`wrangler secret:bulk` is deprecated and will be removed in a future version.

This improves logging in cases where a secret is failing to upload because an environment variable with the same name already exists (see: #240). Example:

Screenshot 2024-11-18 at 9 41 21 AM

With this change, we get this:
Screenshot 2024-11-18 at 9 35 43 AM

### Testing:

Tested with a repo where the Worker already contains a secret of a different type:
https://github.com/jahands/wrangler-action-example/actions/runs/11895527967/job/33145312742

### DIFF
--- a/src/wranglerAction.ts
+++ b/src/wranglerAction.ts
@@ -241,26 +241,6 @@ function getEnvVar(envVar: string) {
 	return value;
 }
 
-async function legacyUploadSecrets(
-	config: WranglerActionConfig,
-	packageManager: PackageManager,
-	secrets: string[],
-	environment?: string,
-	workingDirectory?: string,
-) {
-	for (const secret of secrets) {
-		const args = ["wrangler", "secret", "put", secret];
-		if (environment) {
-			args.push("--env", environment);
-		}
-		await exec(packageManager.exec, args, {
-			cwd: workingDirectory,
-			silent: config["QUIET_MODE"],
-			input: Buffer.from(getSecret(secret)),
-		});
-	}
-}
-
 async function uploadSecrets(
 	config: WranglerActionConfig,
 	packageManager: PackageManager,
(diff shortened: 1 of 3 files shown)


src/wranglerAction.ts
```diff
@@ -241,26 +241,6 @@ function getEnvVar(envVar: string) {
 	return value;
 }
 
-async function legacyUploadSecrets(
```
Maximo-Guk, line 244: Should we keep the legacy upload behavior for people on older versions of wrangler?

Maximo-Guk: Ohh, `"wrangler", "secret:bulk"` is being deprecated, and `"wrangler", "secret", "put"` is the current "legacy" way of doing things

jahands: The legacy behavior _is_ the current behavior :) The "new" behavior was the now-deprecated secret:bulk command.

All this PR does is replace the "non-legacy" behavior with the code from this function

Maximo-Guk (approved):

Maximo-Guk:
We now have some baby tests in `wranglerAction.test.ts` #322 , so it's possible to write some tests for `uploadSecrets()` if you're so inclined!

Maximo-Guk (changes requested):
Based off of @WalshyDev feedback, we should probably use https://github.com/cloudflare/workers-sdk/pull/5988 `wrangler secret bulk` instead, if the user is on a newer version of wrangler where `wrangler secret:bulk` was deprecated

Cherry:
Yeah, this would be a pretty major behavioural change to revert to the old method of setting individual secrets. Using `wrangler secret put` for tens of secrets within a worker can result in errors from the API since they're run in parallel and sometimes don't get acknowledged quickly enough and then cause a mismatch, as well as that many individual new deploys of your worker, cluttering up the versions list.

jahands:
Closing as this was not the right way to fix this. Looking into improving logging in bulk puts

Issue #338, opened 2024-12-11
Wrangler installation fails using pnpm 9.13.2


Leetfs:
```
dependencies:
  + pnpm 9.13.2 (9.15.0 is available)
```

```
Run cloudflare/wrangler-action@v3
🔍 Checking for existing Wrangler installation
📥 Installing Wrangler
  /home/runner/setup-pnpm/node_modules/.bin/pnpm add wrangler@3.[9](https://github.com/project-trans/RLE-wiki/actions/runs/12270487186/job/34235834456#step:8:10)0.0
   ERR_PNPM_ADDING_TO_ROOT  Running this command will add the dependency to the workspace root, which might not be what you want - if you really meant it, make it explicit by running this command again with the -w flag (or --workspace-root). If you don't want to see this warning anymore, you may set the ignore-workspace-root-check setting to true.
Error: The process '/home/runner/setup-pnpm/node_modules/.bin/pnpm' failed with exit code 1
Error: 🚨 Action failed
```

PR #339, opened 2024-12-11
Fix installing Wrangler using pnpm

Leetfs:
issue: #338

test url: https://github.com/Leetfs-dev/test/actions/runs/12271068666/job/34238417617

```
dependencies:
  + pnpm 9.13.2 (9.15.0 is available)
```

```
Run Leetfs/wrangler-action@v1
🔍 Checking for existing Wrangler installation
📥 Installing Wrangler
🚀 Running Wrangler Commands
🏁 Wrangler Action completed
```

### DIFF
--- a/src/packageManagers.ts
+++ b/src/packageManagers.ts
@@ -19,7 +19,7 @@ const PACKAGE_MANAGERS = {
 		execNoInstall: "yarn",
 	},
 	pnpm: {
-		install: "pnpm add",
+		install: "pnpm add -g",
 		exec: "pnpm exec",
 		execNoInstall: "pnpm exec",
 	},


Leetfs:
@Maximo-Guk 

Hi, could you please take a look at this PR when you have time? It's been pending for a while. Let me know if any changes are needed. Thanks.

Maximo-Guk:
Hi @Leetfs 👋, sorry for taking so long to respond!

I'm not sure if it's common for GitHub actions to install things like `pnpm` to the global scope, since it seems to me that now our action might be introducing some unintended side effects.

I think I would recommend users to install pnpm instead using the official pnpm action https://github.com/pnpm/action-setup, and then invoke wrangler-action with your preinstalled pnpm.

What do you think?

Leetfs:
Hi @Maximo-Guk , thanks for your reply

I retested the workflow, using `pnpm/action-setup@v4.1.0` and `cloudflare/wrangler-action@v3.14.0`

Please note:

- pnpm does not allow adding dependencies directly to the workspace root directory by default
- Need to use `-g` `-w` parameters to explicitly specify the installation location

I think it will be more convenient to install `wrangler` to the global path in the workflow, so use the `-g` parameter.

image

image

Maximo-Guk:
@Leetfs Are you able to install wrangler in the root package.json of your project? https://github.com/mtftm/test/blob/main/package.json#L14

If you have wrangler specified in your root package.json, it should become available to the action as well, meaning you wouldn't even need to add `wrangler` to the global scope

Maximo-Guk:
I believe `-w` flag only works when you're inside of a pnpm workspace.

image

I definitely think we should document this better, and urge users to specify `wrangler` in their package.json if they're using a pnpm workspace

Leetfs:
@Maximo-Guk Okay~Thank you for your attention to this PR.

Issue #347, opened 2025-01-20
v3.13.1 causes deployment failure


tenpaiyomi:
No changes introduced outside of the update from v3.13.0 to v3.13.1, however our CI process started throwing an error

```
Error: Resource not accessible by integration - https://docs.github.com/rest/deployments/deployments#create-a-deployment
Error: 🚨 Action failed
```

Applicable line in the Github Action

```
            - name: Deploy to Cloudflare Pages
              uses: cloudflare/wrangler-action@v3
              with:
                  apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
                  accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
                  githubToken: ${{ secrets.GITHUB_TOKEN }}
                  command: pages deploy build/client --project-name ***** --branch ${{ env.GIT_BRANCH }}
                  workingDirectory: apps/web
```

Changing the version back to explicitly `v3.13.0` resolves the issue.

spelexander:
Saw this also

PR #350, opened 2025-02-05
Handle failures in createGitHubDeployment and createGitHubJobSummary

Maximo-Guk:
Addresses #347. 

We shouldn't hard fail the action if we have issues creating the Github deployment and Github job summary.

### DIFF
--- a/src/service/github.ts
+++ b/src/service/github.ts
@@ -1,7 +1,7 @@
 import { summary } from "@actions/core";
 import { context, getOctokit } from "@actions/github";
 import { env } from "process";
-import { info } from "../utils";
+import { info, warn } from "../utils";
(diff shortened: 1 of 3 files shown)


jrf0110 (approved):

src/service/github.ts
```diff
@@ -95,25 +95,34 @@ export async function createGitHubDeploymentAndJobSummary(
...
+							0,
+							8,
+						),
+					deploymentUrl: pagesArtifactFields.url,
+					aliasUrl: pagesArtifactFields.alias,
+				}),
+			]);
+
+		if (createGitHubDeploymentRes.status === "rejected") {
+			info(config, "Creating Github Deployment failed");
```
jrf0110, line 121: maybe be a `warn`?

Issue #343, opened 2025-01-14
`versions` commands do not set `deployment-url`


gferreri:
When using the action with a command to upload a version (and preview URLs are enabled), the action's `deployment-url` output is not set.

### Example Step
```
- name: Deploy Test Report to Cloudflare
        if: always()
        id: deploy-to-cloudflare
        uses: cloudflare/wrangler-action@v3
        with:
          accountId: ${{secrets.CLOUDFLARE_ACCOUNT_ID}}
          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
          workingDirectory: ./apps/my-app
          command: versions upload -c wrangler.test-report.json

 - name: Get deployment URL
        if: always()
        id: get-deployment-url
        run: |
          set -x
          echo "url=${{ steps.deploy-to-cloudflare.outputs.deployment-url  }}" >> "$GITHUB_OUTPUT"
```

### Output
```
Run cloudflare/wrangler-action@v3
/usr/bin/docker exec  804a57670ad65f72e1537be4fdc519444a548e2523b68ab626f5c41ca196ea1d sh -c "cat /etc/*release | grep ^ID"
🔍 Checking for existing Wrangler installation
  /__t/node/20.10.0/x64/bin/npx --no-install wrangler --version
  3.102.0
  ✅ No wrangler version specified, using pre-installed wrangler version 3.102.0

🚀 Running Wrangler Commands
  /__t/node/20.10.0/x64/bin/npx wrangler versions upload -c wrangler.test-report.json
  
   ⛅️ wrangler 3.102.0
  --------------------
  
  🌀 Building list of assets...
  🌀 Starting asset upload...
  🌀 Found 7 new or modified static assets to upload. Proceeding with upload...
  + /data/3a0c226d[11](https://github.com/xxx/External.Frontend/actions/runs/12777465257/job/35618312204#step:13:12)ecbbe01954a623ddbd2e6a9b5e4d3d.webm
  + /index.html
  + /data/4aca4b57288f7e3de5063b3cc15221f42a925f88.zip
  + /data/a606f64154803ab67728381defc4a8a104f1b6a0.webm
  + /data/d8caac46bdd4ff93091c002c783eab03a2e1848b.zip
  + /data/410258399721821f643a023c1644329d1b1e52ad.webm
[...]

Issue #311, opened 2024-10-25
Use wrangler deploy artifacts for deployment URLs


tsteckenborn:
Hi all,
I was wondering why I see some of my environment variables as deployment urls. I would argue that the coding here is not robust enough as I now always get, for example, the posthog endpoint I set as a variable (because it may be us or eu) as my deployment url:

https://github.com/cloudflare/wrangler-action/blob/b2be7abbfc024e06516af00e22260d4b6edcce53/src/index.ts#L343-L349

Would be nice if this could be improved.

nprogers:
This is a bug where wrangler-action is picking up the first url listed in environment variables. We need to create a PR to use deployment metadata instead. Updating this issue to reflect what work needs to be done.

michael-land:
I still have this issue due to it extracts the first URL from stdout, which is `https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/telemetry.md`

verison:
```
    "wrangler": "3.99.0"
    "wrangler-action": "v3.13.0"
```

```
/usr/local/bin/npx wrangler deploy
  Cloudflare collects anonymous telemetry about your usage of Wrangler. Learn more at https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/telemetry.md
   ⛅️ wrangler 3.99.0
...```

**Comment by michael-land** [comment]
In my case, the workaround is to add WRANGLER_SEND_METRICS: false to the environment.

![CleanShot 2024-12-27 at 23 35 44@2x](https://github.com/user-attachments/assets/cac8bc65-887a-40b8-8129-9da1ae05d8f1)

**Comment by iharuya** [comment]
I faced the same issue and this is my workaround to extract the url from stdout:

```
- name: Deploy to Cloudflare
  id: cloudflare_workers_deploy
  uses: cloudflare/wrangler-action@v3
  with:
    apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
    workingDirectory: "apps/backend"

- name: Parse Wrangler Output
  id: parse-wrangler-output
  uses: actions/github-script@v6
  with:
    result-encoding: string
    script: |
      const output = `${{ steps.cloudflare_workers_deploy.outputs.command-output }}`;
      console.log("wrangler output:", output);
      const baseUrl = new URL("https://[worker name].[account name].workers.dev") // your prod url
      const versionIdMatch = output.match(/Current Version ID: ([0-9a-f-]+)/);
[...]

PR #351, opened 2025-02-06
Use wrangler outputs for wrangler deploy and versions upload

Maximo-Guk:
Addresses #343 and #311.

Uses wrangler outputs for wrangler deploy and versions upload so that we can accurately set `deployment-url` for `wrangler deploy` and `versions upload`


src/wranglerAction.ts
```diff
@@ -379,46 +475,23 @@ async function wranglerCommands(
...
-					// DEPRECATED: deployment-alias-url in favour of pages-deployment-alias, drop in next wrangler-action major version change
-					const { deploymentUrl, aliasUrl } =
-						extractDeploymentUrlsFromStdout(stdOut);
-
-					setOutput("deployment-url", deploymentUrl);
-					// DEPRECATED: deployment-alias-url in favour of pages-deployment-alias, drop in next wrangler-action major version change
-					setOutput("deployment-alias-url", aliasUrl);
-					setOutput("pages-deployment-alias-url", aliasUrl);
-				}
+			switch (true) {
```
jrf0110, line 478: Look I kinda like it; But it's kinda clever. It took me a little too long to parse exactly and make sure you were doing what you wanted to do. It would be _much more clear_ with simple `if` statements.

src/wranglerArtifactManager.ts
```diff
@@ -65,34 +90,32 @@ export async function getWranglerArtifacts(
...
  *
  * @param artifactDirectory
- * @returns The first pages-output-detailed found within a wrangler artifact directory
+ * @returns The first SupportedOutputEntry found within a wrangler artifact directory
  */
-export async function getDetailedPagesDeployOutput(
+export async function getOutputEntry(
 	artifactDirectory: string,
-): Promise<OutputEntryPagesDeployment | null> {
+): Promise<SupportedOutputEntry | null> {
```
jrf0110, line 100: We had discussed making this function return an _array of `SupportedOutputEntry`_. This way you can simply loop over the artifacts and apply the outputs based on the artifact type

Maximo-Guk: wrangler-action should only ever have one SupportedOutputEntry, since every time the action is executed it generates a new tmpDir to be used for wranglerArtifacts https://github.com/cloudflare/wrangler-action/blob/main/src/index.ts#L25 .

Also, if it did have multiple SupportedOutputEntry then it would be possible for artifacts to override output from another one, e.g they both set `deployment-url` output

jrf0110 (approved):

PR #355, opened 2025-02-11
feat: :sparkles: added new DEPLOYMENT_NAME input param for the action

VoloshchenkoAl:
The new `DEPLOYMENT_NAME` input parameter was added for the action. This param is responsible for the custom name of the GitHub Deployment

### Why are these changes added?

I have a monorepo with several static sites, that I deploy to the Cloudflare Pages. When I create a Preview for them all, they have the same name — `preview`, and only the last deployed Site is available through the Deployment panel. 

CleanShot 2025-02-12 at 00 28 07@2x

With the changes from this PR, I could:
- set a custom name for my deployments
- all deployments will be available with the deployment panel in the PR

### View words about the previous action

The `pages-action` has another solution [for the problem](https://github.com/cloudflare/pages-action/blob/main/src/index.ts#L142C9-L142C24). I think that the input parameter is a better solution because gives more flexibility for naming


src/index.ts
```diff
@@ -26,7 +26,8 @@ const config: WranglerActionConfig = {
 		tmpdir(),
 		`wranglerArtifacts-${crypto.randomUUID()}`,
 	)}`,
-	GITHUB_TOKEN: getInput("gitHubToken", { required: false }),
+	GITHUB_TOKEN: getInput("gitHubToken"),
```
Maximo-Guk, line 29: I think we still want to leave `GITHUB_TOKEN` as optional for now, otherwise, this will be a breaking change for people who haven't specified a `GITHUB_TOKEN`

src/service/github.ts
```diff
@@ -95,14 +95,15 @@ export async function createGitHubDeploymentAndJobSummary(
 		pagesArtifactFields.deployment_trigger
 	) {
 		const octokit = getOctokit(config.GITHUB_TOKEN);
+		const environment = config.DEPLOYMENT_NAME || pagesArtifactFields.environment;
```
jrencz, line 98: Hi!

First of all: thank you for this PR. I look forward to use it as I use the action to deploy just a documentation of my project and having "production" in deployments bugged me for a while

However: have you considered other name than `DEPLOYMENT_NAME`?

The line I'm pinning a thread to shows that the chosen name might be a bit misleading (deployment name/environment)

I'd suggest `DEPLOYMENT_ENVIRONMENT` as more precise. How about that?

src/index.ts
VoloshchenkoAl: The `required: false` is a default value for `InputOptions` based on the description inside the type file.
CleanShot 2025-04-10 at 22 47 20@2x

But I have returned it to the previous state. Also, I added `{ required: false }` for the new field

src/service/github.ts
VoloshchenkoAl: I have changed the new config field to `DEPLOYMENT_ENVIRONMENT`. Thank you for your suggestion

jrencz (approved):
Thanks for the adjustment. Now it's just right

kubk:
Amazing feature, would be nice to get it merged. My use case is the frontend deployed via Pages to dev and prod workers. Both of them have GitHub deployments with the title 'production', which is confusing

sebdanielsson:
@Maximo-Guk can we get this merged please?👍

WhiteHoodHacker:
Any update from maintainers on getting this merged? I have three sites in a monorepo and quickly getting the preview URLs for each site if they all get updated is slightly annoying :(

jbowring:
+1 vote for this getting merged. Would be a very useful feature, thanks @VoloshchenkoAl for your efforts!

PR #356, opened 2025-02-15
feat: add `version-id` to action output of "version upload" command

iiroj:
This Pull Request adds a new GitHub Action output `version-id` referring to the new version uploaded with `wrangler versions upload`.

The use-case is to be able to run the Wrangler action to upload a new Worker version, and then run it again to deploy that version, for example:

```yaml
      - name: Upload new Worker version
        id: upload
        uses: cloudflare/wrangler-action@v3
        with:
          command: versions upload --tag "${{ github.ref_name }}"
      - name: Deploy new Worker version to 100%
        uses: cloudflare/wrangler-action@v3
        with:
          command: versions deploy "${{ steps.upload.outputs.version-id }}@100%"
```

### DIFF
--- a/src/wranglerArtifactManager.test.ts
+++ b/src/wranglerArtifactManager.test.ts
@@ -1,4 +1,5 @@
 import mockfs from "mock-fs";
+import { randomUUID } from "node:crypto";
 import { afterEach, describe, expect, it } from "vitest";
 import {
 	getOutputEntry,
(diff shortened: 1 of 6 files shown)


src/wranglerArtifactManager.ts
```diff
@@ -45,6 +45,8 @@ export type OutputEntryVersionUpload = z.infer<typeof OutputEntryVersionUpload>;
 const OutputEntryVersionUpload = OutputEntryBase.merge(
 	z.object({
 		type: z.literal("version-upload"),
+		/** The id of this uploaded version. Format is UUID . */
+		version_id: z.string().optional(),
```
iiroj, line 49: This looks like a UUID when running `wrangler` locally with `WRANGLER_OUTPUT_FILE_DIRECTORY` and looking at the output file, but I opted to simply parse it as a string because I assume it's not that important.

Maximo-Guk: I would go ahead and parse it as a UUID since it should only ever be a UUID according to https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/output.ts#L141

if the format changes from UUID, that would definitely be a breaking wrangler change!

iiroj: Sorry, I haven't had the time to get back to this. Is this PR otherwise acceptable if I change this to be parsed as an UUID?

src/wranglerArtifactManager.test.ts
```diff
@@ -154,30 +157,50 @@ describe("wranglerArtifactsManager", () => {
 				expect(artifact).toEqual({
 					version: 1,
 					type: "version-upload",
+					version_id,
 					preview_url: "https://example.com",
 				});
-			}),
-				it("Skips artifact entries that are not parseable", async () => {
```
iiroj, line 160: This diff is slightly larger that necesary because these two were chained with `,` and I refactored them into two separate blocks with `;` and an empty line.

PR #364, opened 2025-04-22
Add wranglerOutputDir input to customize artifact location

horai93:
This PR introduces the optional `wranglerOutputDir` input.

This allows users to specify a custom directory for Wrangler's output artifacts.
The primary motivation is to enable uploading these artifacts using `actions/upload-artifact` in subsequent workflow steps.

### DIFF
--- a/.changeset/rare-ladybugs-ring.md
+++ b/.changeset/rare-ladybugs-ring.md
@@ -1,0 +1,5 @@
+---
+"wrangler-action": patch
+---
+
+Add `wranglerOutputDir` input to customize artifact location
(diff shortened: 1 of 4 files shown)


horai93:
I apologize for the misunderstanding.

This was my mistake. What I actually wanted to do was already possible with:
`command: deploy --outdir ./dist`

I now realize that `WRANGLER_OUTPUT_FILE_DIRECTORY` is not an option for the wrangler deploy command, but rather a different option altogether.
https://github.com/cloudflare/workers-sdk/blob/357d42acfb16d21169d004961030cd4822526a96/packages/wrangler/src/environment-variables/misc-variables.ts#L75-L83

PR #376, opened 2025-06-29
update wrangler version

hanthor:


isaumya:
Instead of specifying the `DEFAULT_WRANGLER_VERSION` to a specific version. The default should always be `latest` flag like `npx`. Cause the wrangler team is always pushing new updates and the action should use the latest action by default.

PR #401, opened 2025-12-23
docs: update README environment parameter examples

curry-battle:
## Summary
 - Update README examples to use `environment` parameter instead of `--env` flag in command
 - This provides clearer guidance on the recommended way to specify environments

 ## Changes
 - Updated workflow dispatch example to use separate `environment` parameter
 - Updated secrets example to use separate `environment` parameter instead of inline `--env` flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,7 +233,8 @@ jobs:
         uses: cloudflare/wrangler-action@v3
         with:
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-          command: deploy --env ${{ github.event.inputs.environment }}
+          command: deploy
+          environment: ${{ github.event.inputs.environment }}  
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -267,7 +268,7 @@ There is an environment parameter that can be set within the workflow to enable
 - uses: cloudflare/wrangler-action@v3
   with:
     apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-    command: deploy --env production
+    command: deploy
     secrets: |
       SUPER_SECRET
     environment: production
```
curry-battle, line 274: While the previous examples with `--env` flag work correctly (the flag takes precedence when both are specified as per [wranglerAction.ts#L321-L323](https://github.com/cloudflare/wrangler-action/blob/707f6375/src/wranglerAction.ts#L321-L323)), using the dedicated `environment` parameter is simpler.

PR #406, opened 2026-03-12
Update default Wrangler version to v4 (4.72.0)

ericclemmons:
The action now defaults to Wrangler v4 instead of v3. All existing version-gated logic (secret upload, deploy vs publish) correctly handles v4. Users can pin to v3 via wranglerVersion input.


ericclemmons:
Closing to reopen from the upstream cloudflare org branch so that CI has access to repository secrets (the fork-based PR runs with 'Secret source: None').

ericclemmons:
Re-opened as https://github.com/cloudflare/wrangler-action/pull/412 (instead of my personal fork)

PR #410, opened 2026-03-18
Upgrade Node.js runtime to v24

JohnTitor:
The Node.js v20 runtime has been deprecated: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
This suppresses the warning on users.

### DIFF
--- a/.changeset/wicked-donuts-fold.md
+++ b/.changeset/wicked-donuts-fold.md
@@ -1,0 +1,5 @@
+---
+"wrangler-action": major
+---
+
+Upgrade Node.js runtime to v24
--- a/.changeset/wicked-donuts-fold.md
+++ b/.changeset/wicked-donuts-fold.md
@@ -1,5 +1,5 @@
 ---
-"wrangler-action": major
+"wrangler-action": patch
 ---
 
 Upgrade Node.js runtime to v24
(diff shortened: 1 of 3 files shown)


kirkouimet:
@JohnTitor thank you!

.changeset/wicked-donuts-fold.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler-action": major
```
ericclemmons: This can be a `patch` since this release is the current active version of Node:
> https://nodejs.org/en/about/previous-releases

See https://developers.cloudflare.com/workers/wrangler/install-and-update/ for supported versions (which technically includes `20`, but this should target the current `lts` that's `25`)


```suggestion
"wrangler-action": patch
```

JohnTitor: @ericclemmons Well, note that it's a technically breaking change on GitHub Actions as it bumps minimum runner version requirement. It's not something about Node.js itself or wrangler at all.
In GHA, such changes are customarily treated as breaking changes, e.g. https://github.com/actions/checkout/releases/tag/v5.0.0

JohnTitor: But yeah, it's up to maintainers to decide whether treating it as breaking or not. I'll follow your way.

JohnTitor:
Seems fixed in https://github.com/cloudflare/wrangler-action/pull/431, @ericclemmons sigh, cloudflare doesn't know open source manner, disappointed.

PR #412, opened 2026-03-24
Update default Wrangler version to v4.x

ericclemmons:
The action now defaults to Wrangler v4 instead of v3. All existing version-gated logic (secret upload, deploy vs publish) correctly handles v4. Users can pin to v3 via wranglerVersion input.


README.md
```diff
@@ -2,6 +2,17 @@
 
 Easy-to-use GitHub Action to use [Wrangler](https://developers.cloudflare.com/workers/cli-wrangler/). Makes deploying Workers a breeze.
 
+## Wrangler v4 Support
```
NuroDev: I'm assuming this is mean't to be titled v3
```suggestion
## Wrangler v3 Support
```

src/index.ts
```diff
@@ -5,7 +5,7 @@ import { getPackageManager } from "./packageManagers";
 import { checkWorkingDirectory } from "./utils";
 import { main, WranglerActionConfig } from "./wranglerAction";
 
-const DEFAULT_WRANGLER_VERSION = "3.90.0";
+const DEFAULT_WRANGLER_VERSION = "4.72.0";
```
ericclemmons: Gotta ask – why isn't this evergreen?

.changeset/wrangler-v4-support.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler-action": minor
```
jahands: Needs to be a major version bump

src/index.ts
jahands: could probably make it use latest v4

ericclemmons: With https://github.com/cloudflare/wrangler-action/pull/425 `4` should work 🙏

.changeset/wrangler-v4-support.md
ericclemmons:
```suggestion
"wrangler-action": major
```

src/index.ts
ericclemmons: Let's see if this can just be `4` now:


```suggestion
const DEFAULT_WRANGLER_VERSION = "4";
```

.changeset/wrangler-v4-support.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler-action": major
+---
+
+Update default Wrangler version to v4 (4.72.0). The action now installs Wrangler v4 by default when no `wranglerVersion` input is specified. Users can still pin to v3 by setting `wranglerVersion: "3.90.0"` explicitly.
```
ericclemmons:
```suggestion
Update default Wrangler version to v4 (`latest`). The action now installs Wrangler v4 by default when no `wranglerVersion` input is specified. Users can still pin to v3 by setting `wranglerVersion: "3.90.0"` explicitly.
```

jahands (approved):

PR #421, opened 2026-04-02
Add getting started guide for new users

### LINKED ISSUE #414: Add documentation on how to get this GitHub action to work from scratch
# Task Description

It would be helpful to add some documentation on how to get this GitHub action to work from scratch. It may look straight-forward to someone who's done it a thousand times, but -- in reality -- it currently isn't.

For example, I am assuming that you need to create an application under `Computer` / `Workers & Pages`. However, this doesn't let you do so. If I select `Create Application` / `Connect GitHub` , it will ask me to approve some repository. I obviously am here, because I don't want this to be handled on Cloudflare's side, but -- rather -- via the GitHub Action. I also don't want it to create a "Hello, World!"-like application because that would deploy some demo app and I already have an `mkdocs` site which I'd like to publish.

So, like... some basic steps to get you set up before you can actually use this GitHub action, would be really helpful. Things like:
* How to create the application
* How to create the secrets / token properly
* How to secure the site with login credentials

# Task Relationships

* cloudflare/workers-sdk#9255

vliggio:
## Summary
- Adds `docs/getting-started.md` with step-by-step instructions for setting up the action from scratch
- Covers: API token creation, finding account ID, adding GitHub secrets, project setup (Workers & Pages), workflow configuration, common issues
- Adds a link from the README to the new guide
- Addresses review feedback: adds Zone Resources permission, clarifies dashboard navigation, fixes Pages auto-creation claims with explicit fallback
- Adds GitHub-hosted vs self-hosted runners section with comparison table and requirements
- Adds stubbed example projects under `docs/examples/` for both Pages and Workers deployments

Closes #414

## Test plan
- [ ] Review guide for accuracy against current Cloudflare dashboard flow
- [ ] Verify the Pages example workflow works for a new project (auto-creation and explicit creation)
- [ ] Verify the Workers example workflow works with a fresh `wrangler init` project
- [ ] Confirm Zone Resources permission is needed for token setup
- [ ] Test example stubs can be copied into a new repo and run successfully

### DIFF
--- a/docs/examples/pages/.github/workflows/deploy.yml
+++ b/docs/examples/pages/.github/workflows/deploy.yml
@@ -1,0 +1,43 @@
+name: Deploy Pages Site
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  deploy:
+    # Use "ubuntu-latest" for a GitHub-hosted runner, or replace with your
+    # self-hosted runner label (e.g. "self-hosted" or [self-hosted, linux]).
+    runs-on: ubuntu-latest
+    name: Deploy to Cloudflare Pages
+    permissions:
+      contents: read
+      deployments: write
+    steps:
+      - uses: actions/checkout@v4
+
+      # Replace this with your actual build step.
+      # The example below is for an mkdocs site — adapt it for
+      # your framework (Next.js, Astro, Hugo, plain HTML, etc.).
+      - name: Build site
+        run: |
+          pip install mkdocs-material
+          mkdocs build
+
+      # If your Pages project does not exist yet, uncomment the step
+      # below to create it before the first deploy.
+      # - name: Create Pages project (first-time only)
+      #   env:
+      #     CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+      #     CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+      #   run: npx wrangler pages project create my-pages-site --production-branch main
+
+      - name: Deploy to Cloudflare Pages
+        uses: cloudflare/wrangler-action@v3
+        with:
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          # Change "site" to your build output directory (e.g. dist/, build/, out/)
+          # Change "my-pages-site" to your desired project name
+          command: pages deploy site --project-name=my-pages-site
--- a/docs/examples/pages/.github/workflows/deploy.yml
+++ b/docs/examples/pages/.github/workflows/deploy.yml
@@ -25,14 +25,7 @@ jobs:
           pip install mkdocs-material
           mkdocs build
 
-      # If your Pages project does not exist yet, uncomment the step
-      # below to create it before the first deploy.
-      # - name: Create Pages project (first-time only)
-      #   env:
-      #     CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-      #     CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
-      #   run: npx wrangler pages project create my-pages-site --production-branch main
-
+      # Note: "pages deploy" auto-creates the Pages project if it doesn't exist yet.
       - name: Deploy to Cloudflare Pages
         uses: cloudflare/wrangler-action@v3
         with:
--- a/docs/examples/worker/.github/workflows/deploy.yml
+++ b/docs/examples/worker/.github/workflows/deploy.yml
@@ -1,0 +1,21 @@
+name: Deploy Worker
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  deploy:
+    # Use "ubuntu-latest" for a GitHub-hosted runner, or replace with your
+    # self-hosted runner label (e.g. "self-hosted" or [self-hosted, linux]).
+    runs-on: ubuntu-latest
+    name: Deploy to Cloudflare Workers
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Deploy to Cloudflare Workers
+        uses: cloudflare/wrangler-action@v3
+        with:
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -1,0 +1,183 @@
+# Getting Started with `wrangler-action` from Scratch
+
+This guide walks you through every step needed to deploy a Cloudflare Workers or Pages project using this GitHub Action, starting from zero.
+
+## Prerequisites
+
+- A [Cloudflare account](https://dash.cloudflare.com/sign-up) (free tier is fine)
+- A GitHub repository containing the project you want to deploy
+
+## Step 1: Create a Cloudflare API Token
+
+The GitHub Action authenticates with Cloudflare using an API token. You must create one with the right permissions.
+
+1. Go to the [Cloudflare dashboard](https://dash.cloudflare.com) and click your profile icon (top right), then **My Profile**.
+2. Select **API Tokens** in the left sidebar.
+3. Click **Create Token**.
+4. Use the **Edit Cloudflare Workers** template (this covers both Workers and Pages). Alternatively, create a custom token with:
+   - **Account** > **Cloudflare Pages** > **Edit**
+   - **Account** > **Workers Scripts** > **Edit**
+   - **Account** > **Account Settings** > **Read**
+   - **Zone** > **Zone** > **Read** (if your Worker uses a custom domain)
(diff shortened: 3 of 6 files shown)


docs/getting-started.md
```diff
@@ -0,0 +1,183 @@
...
+If you don't have a Workers project yet:
+
+```sh
+npx wrangler init my-worker
+cd my-worker
+```
+
+This creates a `wrangler.toml` configuration file and a basic Worker. Push it to your GitHub repo.
+
+Your `wrangler.toml` should include at minimum:
```
carlspring, line 61: Where should this file be placed? Does it need to be passed in as a parameter to the GitHub workflow?

docs/getting-started.md
```diff
@@ -0,0 +1,183 @@
...
+## Step 3: Add Secrets to Your GitHub Repository
+
+Store your Cloudflare credentials as GitHub Actions secrets so they stay encrypted and never appear in logs.
+
+1. In your GitHub repository, go to **Settings** > **Secrets and variables** > **Actions**.
+2. Click **New repository secret** and add:
+   - **Name:** `CLOUDFLARE_API_TOKEN` — **Value:** the API token from Step 1
+   - **Name:** `CLOUDFLARE_ACCOUNT_ID` — **Value:** the account ID from Step 2
+
+## Step 4: Set Up Your Project
```
carlspring, line 46: Are both of the below required, or is it an either-or?

vliggio: Both are required. The token is the key that allows you do things. The account ID is WHERE that key is going to do things (and the key is useless anywhere else).

carlspring: No, I meant does one need to have both workers and pages, or could it be just pages?

vliggio: Ooh, sorry - your comment showed in reference to the API token and accountID. You do not need both - it can be either pages or workers.

carlspring: Thanks for clarifying!

vliggio: Updated doc to clarify either/or in commit fea1419

docs/getting-started.md
vliggio: Updated doc in commit fea1419

docs/getting-started.md
```diff
@@ -0,0 +1,187 @@
...
+2. Select **API Tokens** in the left sidebar.
+3. Click **Create Token**.
+4. Use the **Edit Cloudflare Workers** template (this covers both Workers and Pages). Alternatively, create a custom token with only the permissions you need:
+   - **Account** > **Cloudflare Pages** > **Edit** (if deploying a Pages project)
+   - **Account** > **Workers Scripts** > **Edit** (if deploying a Worker)
+   - **Account** > **Account Settings** > **Read**
+   - **Zone** > **Zone** > **Read** (if your Worker uses a custom domain)
+
+   > You do not need both Pages and Workers permissions — include only what applies to your project.
+5. Under **Account Resources**, select the account you want to deploy to.
```
carlspring, line 25: @vliggio :
From what I have tried, it also requires the "Zone Resources" part here as well.

docs/getting-started.md
```diff
@@ -0,0 +1,187 @@
...
+6. Click **Continue to summary**, then **Create Token**.
+7. **Copy the token** — you won't be able to see it again.
+
+## Step 2: Find Your Cloudflare Account ID
+
+1. Go to the [Cloudflare dashboard](https://dash.cloudflare.com).
+2. Select any zone (domain), or go to **Workers & Pages**.
```
carlspring:
```suggestion
2. Select any zone (domain), or go to **Workers & Pages** under the **Compute** section.
```

docs/getting-started.md
```diff
@@ -0,0 +1,187 @@
...
+compatibility_date = "2024-01-01"
+```
+
+> **Note:** You do _not_ need to set `account_id` in `wrangler.toml` if you pass `accountId` to the GitHub Action (recommended — keeps credentials out of your repo).
+
+### For Cloudflare Pages
+
+If you have a static site (e.g. built with mkdocs, Next.js, Astro, etc.):
+
+1. **You do not need to create a Pages project in the Cloudflare dashboard first.** The `wrangler pages deploy` command will create it for you on the first run.
```
carlspring: Perhaps I'm not doing this right, but I believe this is not true.
I had to actually explicitly add:
```
      - name: Create Cloudflare Pages Project if one does not exist
        env:
          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
        run: |
          if ! npx wrangler@3 pages project list 2>/dev/null | grep -q "my-mkdocs-to-cloudflare-pages"; then
            npx wrangler@3 pages project create my-mkdocs-to-cloudflare-pages --production-branch main
          else
            echo "Cloudflare Pages project 'publish-mkdocs-to-cloudflare-pages' already exists, skipping creation."
          fi
```

Do you think you could create a git repository fully illustrating your steps for a configuration that doesn't use Workers, but just Pages?

docs/getting-started.md
vliggio: Updated that. That's required for zone level configurations. Workers live in the account level, while things like zone specific routes live at the zone level. Good call out.

docs/getting-started.md
vliggio: Adopted

docs/getting-started.md
vliggio: Added two examples, one pages, one workers, and added some other clarifying information for runners.

docs/examples/pages/.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+          pip install mkdocs-material
+          mkdocs build
+
+      # If your Pages project does not exist yet, uncomment the step
+      # below to create it before the first deploy.
+      # - name: Create Pages project (first-time only)
+      #   env:
+      #     CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+      #     CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+      #   run: npx wrangler pages project create my-pages-site --production-branch main
```
carlspring: If I may suggest, something like this will make sure you don't have have to keep commented out code in your workflow (and later wonder if it's dead code):
```suggestion
      - name: Create Cloudflare Pages Project if one does not exist
        env:
          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
        run: |
          if ! npx wrangler@3 pages project list 2>/dev/null | grep -q "my-pages-site"; then
            npx wrangler@3 pages project create my-pages-site --production-branch main
          else
            echo "Cloudflare Pages project 'my-pages-site' already exists, skipping creation."
          fi
```

I'm not sure if the CLI is intended to actually create the project for you upon the first run, but in case that wasn't by design, you could always have the check above to make it bulletproof.

docs/examples/worker/.github/workflows/deploy.yml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  deploy:
+    # Use "ubuntu-latest" for a GitHub-hosted runner, or replace with your
+    # self-hosted runner label (e.g. "self-hosted" or [self-hosted, linux]).
+    runs-on: ubuntu-latest
+    name: Deploy to Cloudflare Workers
+    steps:
+      - uses: actions/checkout@v4
+
```
carlspring, line 16: Not sure if the check above is required here as well...?

docs/getting-started.md
vliggio: Good call on avoiding commented-out code. Since `wrangler pages deploy` auto-creates the project if it doesn't already exist, I've removed the commented block entirely and replaced it with a one-line note. No manual creation step needed.

docs/examples/worker/.github/workflows/deploy.yml
vliggio: Not needed here — `wrangler deploy` for Workers automatically creates the Worker if it doesn't exist, as long as there's a valid `wrangler.toml`. There's also no `workers project create` CLI equivalent like there is for Pages.

docs/examples/pages/.github/workflows/deploy.yml
vliggio: The suggestion to replace the commented-out block with an idempotent if ! ... grep check is a nice improvement avoiding dead-looking commented code and is safe to leave in the workflow permanently. That said, wrangler pages deploy will auto-create the project if it doesn't exist, so this step is technically optional. Two choices:
  - replace the commented block with the idempotent version OR
  - remove the commented block entirely and add a one-line comment noting auto-creation.
Simpler is often better, so took the second path and removed it altogether.

docs/getting-started.md
carlspring: This is very strange. It doesn't actually create it for me! Are you sure? Are the docs for the CLI mentioning this anywhere? It may be the case that I am somehow hitting cloudflare/workers-sdk#9255 ...?

carlspring:
@ascorbic / @cloudflare/wrangler @cloudflare/banda-team,

Ping?

carlspring:
Just checking -- will this ever get merged?

PR #423, opened 2026-04-14
Enhance README with Wrangler configuration details

ckhicks:
Added sections for wranglerVersion, workingDirectory, Secrets, and Commands to the README.

I noticed old links - both from support engineers and in the community - that directed people to the "wranglerVersion" heading. This commit adds that back (along with a few others) to ensure that those links resolve properly in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,8 @@ jobs:
 
 ## Configuration
 
+### wranglerVersion
+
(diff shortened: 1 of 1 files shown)


ckhicks:
We may also want to add some additional definition around the Wrangler Version option, specifically. It's not clearly documented anywhere that the action maintains conservative defaults to avoid breaking existing workflows when Wrangler introduces breaking changes. While this prevents unexpected failures, it means new features require explicit declaration in the configuration file.

ckhicks:
Changes now included/deprecated by #427

Issue #379, opened 2025-07-08
No planning to upgrade to fully support version 4+?


BaseMax:
No planning to upgrade to fully support version 4+?

kubk:
Bump so the maintainer could see

ToshY:
First time using workers I tried using [cloudflare/templates](https://github.com/cloudflare/templates), specifically [chanfana-openapi-template](https://github.com/cloudflare/templates/tree/main/chanfana-openapi-template). These templates use wrangler `v4.21.x` as dev dependency and the `wrangler.jsonc` file, which is supported starting from `v3.91.0`.

Without much thought, I've used the [Github Actions](https://developers.cloudflare.com/workers/ci-cd/external-cicd/github-actions/#github-actions) example in the worker documentation and thought it would be fine. Not only is it still using `v3.90.0` and therefore not supporting the `.json` / `.jsonc` extension, there is apparently also no `v4` of this action.

<details><summary>action logs</summary>

```text
  /usr/local/bin/npx wrangler deploy
  
   ⛅️ wrangler 3.90.0 (update available 4.34.0)
  ---------------------------------------------
  
  ▲ [WARNING] The version of Wrangler you are using is now out-of-date.
  
    Please update to the latest version to prevent critical errors.
    Run `npm install --save-dev wrangler@4` to update to the latest version.
    After installation, run Wrangler with `npx wrangler`.
  
  
  
  ✘ [ERROR] Missing entry-point: The entry-point should be specified via the command line (e.g. `wrangler deploy path/to/script`) or the `main` config field.
```

</details>

@Maximo-Guk

There are already multiple PRs trying to address the same problem (#363, #376), but by the looks of it the action doesn't seem to be very well maintained anymore.

Can we expect new releases anytime soon?

---

**TLDR**

I'm using [`wranglerVersion`](https://github.com/cloudflare/wrangler-action?tab=readme-ov-file#configuration) for the time being:

```yaml
      - name: Build & Deploy Worker
        uses: cloudflare/wrangler-action@v3
        with:
          wranglerVersion: '4.34.0'
          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
```

Ruben-Smit:
I didn't really try the above solution but this also worked for me:

```yaml
  deploy:
    runs-on: ubuntu-latest
    name: Deploy
    steps:
[...]

Issue #366, opened 2025-05-02
Request: Install Newer Version of Wrangler by Default


CNX-Inc:
Noting, this might be something with my project (I'm not much of a node user but i have updated my `package.json` and `package-lock.json` files to no avail.

It seems as though the action installs an outdated wrangler by default, from my logs:
 `/usr/local/bin/npm i wrangler@3.90.0`

Currently I'm using Wrangler 4.14.1 (locally) and the older version seems to not like to deploy.

Here's how I've worked around the issue, but thought you all might want to update the default or maybe provide an `version: 4.14.1` option to `with:` 

Thanks!

```jobs:
  deploy:
    runs-on: ubuntu-latest
    name: Deploy
    steps:
      - uses: actions/checkout@v4
      - name: Deploy
        uses: cloudflare/wrangler-action@v3
        with:
          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
          workingDirectory: "data-worker"
          preCommands: |
            echo "*** pre commands ***"
            echo `pwd`
            npx wrangler --version
            npm install wrangler@latest
            echo "******"
```

mamiu:
Had the same problem right now when trying to use the IMAGES binding.

RobFaustLZ:
Following. All the Wrangler project initialization scripts have been updated to use `.jsonc` config files by default rather than the old `.toml` format. But, `.jsonc` isn't supported until Wrangler 3.91 and the default is currently set to 3.90.

omkhar:
I presume there is a valid reason for not changing default behavior. Rather than invoking a manual installation, you can just pass `wranglerVersion` like this:
      
```
     - uses: actions/checkout@v4
      - uses: cloudflare/wrangler-action@v3
        name: Deploy to Cloudflare (Production)
        with:
          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
          wranglerVersion: "latest"
```

CNX-Inc:
@omkhar  - are you suggesting `wranglerVersion: "latest"` will install the latest? (vs. my hack install)? 

That seems cleaner for sure.

stof:
An alternative is to add a step running `npm ci` before the step using `cloudflare/wrangler-action`, to install the npm dependencies of the project. In such case, `cloudflare/wrangler-action` will use the version of wrangler installed in your project.

[...]

PR #426, opened 2026-04-15
Support version ranges and tags in wranglerVersion input

### LINKED ISSUE #390: cannot use wranglerVersion to use latest v4 release
It crashes in uploadSecrets() https://github.com/cloudflare/wrangler-action/blob/707f63750981584eb6abc365a50d441516fb04b8/src/wranglerAction.ts#L259-L273 with `Error: Invalid Version:` which appears to be a bug.

These all fail:
```yaml
wranglerVersion: 4
wranglerVersion: v4
wranglerVersion: 4.*
wranglerVersion: 4.x
wranglerVersion: 4.*.*
```

We don't want to hard-code the full version. Nor do we want "latest" as that might bump to v5 someday. We only want the latest v4.

Workaround was to avoid b0rk wranglerVersion and add this step before cloudflare/wrangler-action.
```yaml
      - name: Install Wrangler v4
        run: npm install --save-dev wrangler@4
```

WillTaylorDev:
The wranglerVersion input only accepted exact versions like '4.81.0'. Setting it to a major version like '4', a range like '^4.0.0', or a tag like 'latest' crashed with 'Invalid Version' during secret uploads. npm installed the right version fine, but subsequent version comparisons failed because they expected exact X.Y.Z format.

After installing wrangler, the action now runs wrangler --version to get the concrete installed version and uses that for all version comparisons. If version detection fails and the user gave an exact version, the action falls back to using it directly, preserving existing behavior.

Pre-installed versions that satisfy a range skip reinstallation.

Fixes #390
Relates to #366
Relates to #379

### DIFF
--- a/.changeset/support-version-ranges.md
+++ b/.changeset/support-version-ranges.md
@@ -1,0 +1,5 @@
+---
+"wrangler-action": minor
+---
+
+Support version ranges and tags in `wranglerVersion` input. You can now set `wranglerVersion` to values like `4`, `^4.0.0`, `4.x`, or `latest` instead of only exact versions like `4.81.0`.
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ jobs:
 
 ## Configuration
 
-If you need to install a specific version of Wrangler to use for deployment, you can also pass the input `wranglerVersion` to install a specific version of Wrangler from NPM. This should be a [SemVer](https://semver.org/)-style version number, such as `2.20.0`:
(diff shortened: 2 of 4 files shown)


ericclemmons (approved):
I've said it once in https://github.com/cloudflare/wrangler-action/pull/425 and I'll say it again, 

> Much better DX for `wranglerVersion`! Good work!

https://github.com/cloudflare/wrangler-action/pull/412 needs this to be evergreen on v4.

Issue #446, opened 2026-09-17
Upstream wrangler action secrets bug


hectorromerodev:
### What happened

`secrets:` is implemented as `wrangler secret bulk`, which resolves the target Worker from the
`name` field in `wrangler.toml`. It ignores a `--name` override passed in `command:`.

So this config, intended to deploy an ephemeral per-PR preview, deploys the preview correctly **and writes its secrets onto the production Worker**:

```yaml
- uses: cloudflare/wrangler-action@v3
  with:
    apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
    command: deploy --name my-worker-pr-123
    secrets: |
      TURNSTILE_SECRET
  env:
    TURNSTILE_SECRET: ${{ secrets.PREVIEW_TURNSTILE_SECRET }}
```

`deploy --name my-worker-pr-123` → creates/updates `my-worker-pr-123`. ✅
`secrets:` → writes to `my-worker` (the `wrangler.toml` name). ❌

### Why it matters

Preview environments legitimately carry dummy or sandbox credentials. On a real site this silently replaced production's captcha secret and mail-provider key with the preview's dummy values
on every PR deploy. The deploy step reports success, the preview works, and production breaks with no error anywhere (the secrets are write-only, so nothing surfaces until a user hits the broken path).

It took a live outage of a contact form and an AI chat endpoint to find.

### Expected

Either:
1. `secrets:` honors the `--name` present in `command:`, or
2. the action **fails loudly** when `command:` contains `--name` and `secrets:` is set, since it cannot honor it.

Option 2 alone would be enough. Silently targeting a different Worker than the one just deployed is the dangerous part.

### Workaround

Drop `secrets:` and push them in a separate step that does honor `--name`:

```yaml
- run: |
    printf '%s' "$TURNSTILE_SECRET" \
      | bunx wrangler secret put TURNSTILE_SECRET --name "my-worker-pr-123"
```

### Docs

The `secrets:` input is documented without mentioning that it ignores `--name`, which is what made the behavior surprising rather than merely undocumented.

PR #447, opened 2026-09-17
feat: add workerName input for named Worker secrets

cjol:
This PR adds a structured `workerName` action input that applies to managed secret uploads and Wrangler commands which accept `--name`. It also rejects command-level `--name` when `secrets` is configured, preventing the mismatched targets reported in #446.

## Why

- The `secrets` input uploads secrets before running the configured command, but `command: deploy --name preview-worker` only applies the name to the deployment.
- Secret upload then falls back to the Worker named in `wrangler.toml`, which can silently replace production secrets while deploying a different preview Worker.
- Parsing an arbitrary command string to copy the value into a second invocation would only approximate command-line argument semantics.
- A structured input follows the existing `environment` pattern: the action owns the shared target and forwards it to every relevant Wrangler invocation.

## Public API Surface

| Input | Kind | Notes |
| --- | --- | --- |
| `workerName` | Optional action input | Adds `--name` to managed secret uploads and supported Worker-targeting commands. |

Supported commands are defined in one command-prefix list and currently include deploy, publish, dev, preview, rollback, deployments, secrets, versions, and trigger deployment operations.

When `secrets` is configured, putting `--name` in `command` now fails with a message directing users to `workerName`. Workflows using `--name` without `secrets` keep their current behavior.

## Code Changes

- Read and validate the new `workerName` input through the action configuration.
- Forward `workerName` to legacy and current managed secret uploads.
- Keep a single extensible list of Wrangler commands that accept `--name`, and apply `workerName` when a configured command matches one of those prefixes.
- Detect command-level `--name` and `--name=value` without extracting or interpreting their values.
- Exercise named deployment plus secret upload against a live Worker in CI, verify the exact secret values through its health endpoint, and always clean it up.
- Document the input and add a patch changeset.

## Compatibility

Workflows that combine `secrets` with `command: deploy --name ...` or another named Wrangler command must move the name into the new input:

```yaml
with:
  workerName: my-worker-pr-123
  command: deploy
  secrets: |
    TURNSTILE_SECRET
```

This intentionally replaces a configuration that could run against one Worker while uploading secrets to another with an explicit error and migration path.


action.yml
```diff
@@ -29,6 +29,9 @@ inputs:
   secrets:
     description: "A string of environment variable names, separated by newlines. These will be bound to your Worker as Secrets and must match the names of environment variables declared in `env` of this workflow."
     required: false
+  workerName:
+    description: "The name of the Worker to deploy or publish and bind secrets to. Use this instead of passing `--name` in `command` when secrets are configured."
```
cjol:
```suggestion
    description: "The name of the Worker to deploy or publish and bind secrets to.
```

README.md
```diff
@@ -89,7 +89,7 @@ jobs:
         workingDirectory: "subfoldername"
 ```
 
-[Worker secrets](https://developers.cloudflare.com/workers/tooling/wrangler/secrets/) can optionally be passed in via `secrets` as a string of names separated by newlines. Each secret name must match the name of an environment variable specified in the `env` field. This creates or replaces the value for the Worker secret using the `wrangler secret put` command. It's also possible to specify worker environment using environment parameter.
+[Worker secrets](https://developers.cloudflare.com/workers/tooling/wrangler/secrets/) can optionally be passed in via `secrets` as a string of names separated by newlines. Each secret name must match the name of an environment variable specified in the `env` field. This creates or replaces the value for the Worker secret using Wrangler's secret commands. Use the `workerName` input to deploy or publish a named Worker and upload secrets to that same Worker. When secrets are configured, passing `--name` through `command` fails with an error directing you to `workerName`. It's also possible to specify a Worker environment using the `environment` input.
```
cjol:
```suggestion
[Worker secrets](https://developers.cloudflare.com/workers/tooling/wrangler/secrets/) can optionally be passed in via `secrets` as a string of names separated by newlines. Each secret name must match the name of an environment variable specified in the `env` field. This creates or replaces the value for the Worker secret using Wrangler's secret commands. Use the `workerName` input to deploy or publish a named Worker and upload secrets to that same Worker. It's also possible to specify a Worker environment using the `environment` input.
```

src/wranglerArtifactManager.test.ts
NuroDev, line 1: Why do these tests need wrapping in brackets?

.github/workflows/deploy.yml
```diff
@@ -95,14 +95,34 @@ jobs:
...
+        env:
+          SECRET1: ${{ secrets.SECRET1 }}
+          SECRET2: ${{ secrets.SECRET2 }}
+
+      - name: Health Check Named Worker
+        run: node .github/workflows/workerHealthCheck.ts wrangler-action-test-secrets-named
+        shell: bash
+
       - name: Clean Up Deployed Workers
+        if: always()
```
NuroDev, line 117: Why is this needed?

src/wranglerArtifactManager.test.ts
cjol: Prettier auto-fixed it badly and I didn’t notice because it wasn’t my change! I’ve split out the prettier fixes into #448 (apparently no one reran prettier since the last update)

.github/workflows/deploy.yml
cjol: without it, if one deployment succeeded but another failed, we might never clean up the successful one

PR #450, opened 2026-09-21
feat: add Workers Previews support

podonnell-dev:
Adds `wrangler preview` support to the v4 action:

- Parses Preview artifacts and exposes stable Preview outputs
- Creates non-production GitHub Deployments and job summaries
- Supports Preview `--var` injection and stdout fallback
- Adds documentation, tests, changeset, and gated live validation via `WORKERS_PREVIEWS_LIVE`

I'm unable to run the E2E test suite on a fork 😞  but local validation against binary mostly worked! I ended up catching a bug in our SDK response - we do not return worker_name (which this expects) so companion wrangler sdk PR is up: https://github.com/cloudflare/workers-sdk/pull/15761

**We'll want this to go in only after that is fixed**


.github/workflows/deploy.yml
```diff
@@ -105,6 +111,33 @@ jobs:
           postCommands: npx wrangler delete --name wrangler-action-test-secrets-default --force
       # END Setup and teardown of Workers w/ Secrets Tests
 
+      # START Setup and teardown of Workers Preview Tests
+      - name: Deploy Workers Preview
+        if: vars.WORKERS_PREVIEWS_LIVE == 'true'
```
ericclemmons: This doesn't need the same security check?
```yaml
        if: github.event.pull_request.head.repo.full_name == github.repository
```

ericclemmons: (Or is it a non-issue because the steps in `uses` have that check? 🤔 )

src/service/github.spec.ts
```diff
@@ -1,14 +1,24 @@
...
+	createPreviewGitHubDeploymentAndJobSummary,
+} from "./github";
+import * as actionsGithub from "@actions/github";
 import { getOctokit } from "@actions/github";
 import { mockGithubDeployments } from "../test/mocks";
 import { getTestConfig } from "../test/test-utils";
 import mockfs from "mock-fs";
 import { readFile } from "fs/promises";
 
+const createOctokit = getOctokit;
```
ericclemmons: Unsure why this is aliased here (vs. on the import, or at all)?

src/wranglerArtifactManager.test.ts
```diff
@@ -38,7 +38,7 @@ describe("wranglerArtifactsManager", () => {
 
 	describe("getOutputEntry()", async () => {
 		describe("OutputEntryPagesDeployment", async () => {
-			it("Returns only detailed pages deploy output from wrangler artifacts", async () => {
+			(it("Returns only detailed pages deploy output from wrangler artifacts", async () => {
```
ericclemmons: I'm not following why this & the other changes needed `it` wrapped in `(...)`? Is it an immediately-invoked function?

ericclemmons (commented):
Just had a couple questions before the lgtm stamp 🙏

.github/workflows/deploy.yml
podonnell-dev: Not a security concern since this repo doesn't allow fork access to secrets but good call on convention mismatch. I'll match the others guards

src/wranglerArtifactManager.test.ts
podonnell-dev: addressed 5e646265cbeeac8891b70571be3b190627bb4ea3
three tests were running inside this comma expression, separated them out

src/service/github.spec.ts
podonnell-dev: unsure too 🤔 just missed this. addressed https://github.com/cloudflare/wrangler-action/commit/5e646265cbeeac8891b70571be3b190627bb4ea3

ericclemmons (approved):
Thanks for addressing my comments ❤️

PR #453, opened 2026-09-22
fix: Publish compiled tags before immutable releases

ericclemmons:
When releasing [v4.1.1](https://github.com/cloudflare/wrangler-action/releases/tag/v4.1.1), we ran into an error:
> Cannot update this protected ref.
> – https://github.com/cloudflare/wrangler-action/actions/runs/35761013548/job/106858767587

v4.1.0 enabled `immutable: true` for GitHub releases.

This PR makes it so that:
- Changesets bumps the version locally (no release)
- Tag gets pushed (so that `build-and-tag-action` requires a remote tag)
- Action builds `dist/index.mjs` as usual
- Verify that tags contain `dist/index.mjs`
- _Finally_, create the **immutable GitHub release**.

`v4` is a "floating" tag since it moves with the latest patch.


.github/workflows/release.yml
```diff
@@ -50,6 +59,35 @@ jobs:
...
+        if: steps.changesets.outputs.published == 'true'
+        env:
+          RELEASE_TAG: ${{ steps.release.outputs.tag }}
+        run: |
+          MAJOR_TAG="${RELEASE_TAG%%.*}"
+          git fetch --force origin \
+            "refs/tags/${RELEASE_TAG}:refs/tags/${RELEASE_TAG}" \
+            "refs/tags/${MAJOR_TAG}:refs/tags/${MAJOR_TAG}"
+          git cat-file -e "${RELEASE_TAG}:dist/index.mjs"
+          git cat-file -e "${MAJOR_TAG}:dist/index.mjs"
```
podonnell-dev: only gap between mine is here - could also assert both tags resolve to the same commit:

`test "$(git rev-parse "${RELEASE_TAG}^{commit}")" = "$(git rev-parse "${MAJOR_TAG}^{commit}")"`

ericclemmons: Added in [ecefb49](https://github.com/cloudflare/wrangler-action/pull/453/commits/ecefb490ff94b0d36197a49f44eb3bb3e303a2a6) that builds on [7e340a9](https://github.com/cloudflare/wrangler-action/pull/453/commits/7e340a9ca944e149d082eeb4ffb5c87b89706db8)

scripts/release.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+npm run build
+git add --force dist/index.mjs
+git commit --message "Build action"
+npx changeset tag
```
podonnell-dev: thinking about it more, since `npx changeset tag` publishes an immutable release, it's probably better to nix the `Push and verify dist-tags` job all together and perform those operations in here prior to this line. So adding:
```
MAJOR_TAG="${RELEASE_TAG%%.*}"
git push --force origin "HEAD:refs/tags/${MAJOR_TAG}"
git fetch --force origin "refs/tags/${MAJOR_TAG}:refs/tags/${MAJOR_TAG}"
git cat-file -e "${MAJOR_TAG}:dist/index.mjs"
test "$(git rev-parse HEAD)" = "$(git rev-parse "${MAJOR_TAG}^{commit}")"

npx changeset tag
```

podonnell-dev: thinking this makes the workflow retryable

ericclemmons: Agreed. Let's see if [d7fd981](https://github.com/cloudflare/wrangler-action/pull/453/commits/d7fd981a11ac968c6a24529a5e4a044ab0779ab4) does it

scripts/release.sh
dario-piotrowicz: I think that some code comments in this file would really help to clarify what it's doing and more importantly why

dario-piotrowicz (approved):

podonnell-dev (approved):

scripts/release.sh
ericclemmons: Great suggestion! Especially since this _should work_ but the intended flow is understood.

Added comments in [5fa9855](https://github.com/cloudflare/wrangler-action/pull/453/commits/5fa985557e6854cc247573d65d37eef978bd0daf)

